PR #11, opened 2019-05-23
Add bundler.rbi from pay-server

jez:


jez:
Also: I disabled the "dismiss review" feature to match how it is in sorbet.

lgtm

PR #13, opened 2019-05-25
Add Stripe::Card with a few fields

frewsxcv:


ptarjan:
Thanks for the contribution! Have you tested using these internally with your app?

frewsxcv:
Yep, I've done some basic testing with it. In general though, the Ruby Stripe gem is a little special in that all the main types don't inherently have any attributes until the underlying data is set (via [this](https://github.com/stripe/stripe-ruby/blob/ec91de6849f34d8d6701a6e91a1b2ee0d50c21ea/lib/stripe/stripe_object.rb#L368)):

```ruby
[4] pry(main)> Stripe::Card.construct_from(brand: 'Visa').brand
=> "Visa"
[5] pry(main)> Stripe::Card.new.brand
NoMethodError: undefined method `brand' for "#<Stripe::Card:0x3fdb27ae6e50> JSON: {\n}"
from /Users/coreyfarwell/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/stripe-4.17.0/lib/stripe/stripe_object.rb:386:in `method_missing'
```

ptarjan:
Ouch, yeah.. Makes sense. I agree with your choice here to code for the common case. Sadly we won't catch non-defensive code with unpopulated objects but we can work on that later, better to get the common case now.

PR #15, opened 2019-05-25
Update state_machines.rbi

frewsxcv:
Necessary to get basic state_machines functionality. e.g.

```ruby
class MyModel
  state_machine :state, initial: 'unauthorized' do
    event :authorize do
      transition 'unauthorized' => 'authorized'
    end
  end
end
```


frewsxcv:
https://github.com/state-machines/state_machines

PR #23, opened 2019-05-31
Add sigs for lhm

pje:
This changeset adds sigs for the majority of the current public API for [LHM](https://github.com/soundcloud/lhm), at `~>2.2`.

🙇 🍨 🚀


ptarjan:
Thanks! Is there anything special to 2.2 for this? If not, you can move it to `all`. Generally we've been using the version ones for one-off-y stuff. I'll merge anyway and let you think about that.

pje:
Thanks! No, nothing special to 2.2; this could be in `all`.

PR #27, opened 2019-06-03
Fix type definitions in ActiveRecord::Base

frewsxcv:
I think I've managed to introduce bad type definitions in every PR I've opened so far 🙃 🏆 

cc @manhhung741


ptarjan:
:) You can test them on your local branch by just putting them anywhere in your tree and it should be the same as upstreaming them.

PR #30, opened 2019-06-04
🤦‍ `try` accepts an *optional* block

pje:
After #29.


ptarjan:
You can test these locally by just putting them in your directory, they should be picked up

PR #32, opened 2019-06-20
Create rbi file for relative_time

davydovanton:
Hey!

Thanks for the library. I played with sorbet in my personal project and created a rbi file for my small library ( https://github.com/davydovanton/relative_time). I will be happy to add it to the main rbi file repository.

Also, I have question. If I understand right you want to use this repository as a registry for all ruby libraries. And everyone can add own lib type annotation file. Is it true?

### DIFF
--- a/lib/relative_time/all/relative_time.rbi
+++ b/lib/relative_time/all/relative_time.rbi
@@ -1,0 +1,34 @@
+# typed: strong
+
+module RelativeTime
+  sig do
+    params(
+      date_to: T.any(Time, DateTime, Date),
+      date_from: T.any(NilClass, Time, DateTime, Date),
+      locale: T.any(NilClass, String, Symbol)
+    ).returns(String)
+  end
+  def self.in_words(date_to, date_from = nil, locale: nil); end
+end
+
+class RelativeTime::InWords
+  sig do
+    params(
+      date_to: T.any(Time, DateTime, Date),
+      date_from: T.any(Time, DateTime, Date)
+    ).returns(String)
+  end
+  def call(date_to, date_from); end
+
+  sig do
+    params(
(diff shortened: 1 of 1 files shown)


davydovanton:
Also, one more question. What to do with different version of one gem (something like rails3 and rails6)?

ptarjan:
If you are the library owner, may I suggest you add a `rbi/` directory to your `gem`? Then you don't need to have a file here. https://sorbet.org/docs/rbi#rbis-within-gems

davydovanton:
@ptarjan hey, thanks! I found this way in the documentation and tried to add `rbis/` folder to my library (also I tried it with `rbi/`):
https://github.com/davydovanton/relative_time/tree/master/rbis

After that I used `github` and `path` link in my gemfile (`gem 'relative_time', github: '...'` or `path: '/users/anton/,..../relative_time'`) for library. After that I called `srb rbi sorbet-typed` and `srb rbi update` and see nothing in my `project/sorbet/` folder. I mean that `rbi_list/` folder was not created. What should I do in this case? I can send any specific information if you want or create a simple repository for reproducing it.

I really want to understand how to include rbi files to my libraries and maybe help with documentation section for library owners.

ptarjan:
You actually won't see any artifact in your directory, but please check end-to-end that sorbet understands another project that has your gem in their Gemfile. This is because we put the paths in the `/Users/pt/.cache/sorbet/gem-rbis/<some_hash>` file (because it is local to your machine, we don't want it committed to your source control).

davydovanton:
@ptarjan thanks a lot for your help! ❤️ 

I'll be happy to add it to the documentation, WDYT?

ptarjan:
Please do!

PR #41, opened 2019-06-28
Type PaperTrail

AlexEvanczuk:
Starting types for https://github.com/paper-trail-gem/paper_trail

I am OOO until July 10th, but feel free to edit and/or merge if you'd like before I get back.  Otherwise, I'll incorporate any requested changes when I return.


DarkDimius:
Thank you, Alex!

PR #43, opened 2019-07-06
Add Travis CI config and script to typecheck

frewsxcv:
Thoughts? I accidentally added a couple malformed RBIs to this repo in the past and this script would have caught the issues 😅


connorshea:
Bundler should probably be fixed and removed from the exclusion list (though that should be done in a separate PR)

frewsxcv:
@aisamanra FYI, for this to work, you (or someone from Stripe) will need to enable this repo in Travis CI

PR #50, opened 2019-07-11
Add T::Boolean to ActiveRecord's rbi for polymorphic support

astrocket:
Fix to support polymorphic

```bash
Expected T.nilable(T.any(Symbol, String)) but found TrueClass for argument polymorphic https://srb.help/7002
     4 |  belongs_to :item, polymorphic: true
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -128,7 +128,7 @@ module ActiveRecord::Associations::ClassMethods
       foreign_type: T.nilable(T.any(Symbol, String)),
       inverse_of: T.nilable(T.any(Symbol, String)),
       optional: T.nilable(T::Boolean),
-      polymorphic: T.nilable(T.any(Symbol, String)),
+      polymorphic: T.nilable(T.any(Symbol, String, T::Boolean)),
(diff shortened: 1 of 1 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -128,7 +128,7 @@ module ActiveRecord::Associations::ClassMethods
       foreign_type: T.nilable(T.any(Symbol, String)),
       inverse_of: T.nilable(T.any(Symbol, String)),
       optional: T.nilable(T::Boolean),
-      polymorphic: T.nilable(T.any(Symbol, String)),
+      polymorphic: T.nilable(T.any(Symbol, String, T::Boolean)),
```
pje: Thanks for catching this! 

I actually think it only needs to allow Boolean values; Strings or Symbols don't really make sense here, AFAIK.

```suggestion
      polymorphic: T.nilable(T::Boolean),
```

astrocket: You are right.
I fixed it !

aisamanra:
Thank you!

PR #51, opened 2019-07-11
Support `uniqueness` in ActiveRecord's `validates`

phansch:
Based on https://api.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of


aisamanra:
Thank you!

PR #53, opened 2019-07-11
Add sidekiq_options type to Sidekiq::Worker::ClassMethods

AlexEvanczuk:
Adds new method for classes that use sidekiq.


aisamanra:
Thank you!

PR #54, opened 2019-07-12
Improve sigs for ActiveRecord `validate`

pje:
Fixes a couple issues:

1) `:presence, :inclusion, :allow_nil` aren't allowed keywords for `validate`—they're only allowed in `validates.`

```ruby
>> class Foo < ActiveRecord::Base; validate :bar, presence: true; end
ArgumentError: Unknown key: :presence. Valid keys are: :on, :if, :unless, :prepend. Perhaps you meant to call `validates` instead of `validate`?
```

2) `validate` takes positional args as a splat, which sorbet [doesn't fully handle yet](https://sorbet.org/docs/error-reference#7019). Notably it doesn't actually accept an array as a first arg. That would throw an error: 

```ruby
>> class Foo < ActiveRecord::Base; validate([:bar, :baz]); end
>> Foo.new.valid?
NoMethodError: undefined method `validate' for [:bar, :baz]:Array
```

~~3) None of the optional keyword args actually need to be `nilable`—they have default values in the method definition itself, so sorbet will allow them to be omitted. I'd argue we don't actually want to allow `nil` to be passed in as a value, so I made them all non-nilable.~~

**EDIT** I was wrong about this! [Sorbet.run doesn't allow this syntax](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params(bar%3A%20String).void%20%7D%0Adef%20foo(bar%3A%20nil)%3B%20end%0A%0Afoo()%0A), but my version of Sorbet (`0.4.4429`) _does_. Is this a sorbet bug?

### DIFF
--- a/lib/activemodel/all/activemodel.rbi
+++ b/lib/activemodel/all/activemodel.rbi
@@ -25,6 +25,7 @@ module ActiveModel::Validations
         on: T.any(Symbol, String),
         prepend: T::Boolean,
         unless: T.any(Symbol, String, T.proc.params(arg0: T.untyped).returns(T::Boolean)),
+        blk: T.nilable(T.proc.params(arg0: T.untyped).void)
       ).void
     end
     def validate(
--- a/lib/activemodel/all/activemodel.rbi
+++ b/lib/activemodel/all/activemodel.rbi
@@ -29,10 +29,11 @@ module ActiveModel::Validations
     end
     def validate(
       *names,
-      if: nil,
-      on: nil,
-      prepend: T::Boolean,
-      unless: nil
+      if: :_,
+      on: :_,
+      prepend: false,
+      unless: :_,
+      &blk
     ); end
 
     # https://github.com/rails/rails/blob/v5.2.3/activemodel/lib/active_model/validations/validates.rb#L75-L105


pje:
cc @AlexEvanczuk: thanks for originally adding this sig in #49!

lib/activerecord/all/activerecord.rbi
```diff
@@ -493,30 +493,23 @@ class ActiveRecord::Base
...
-    presence: nil,
-    inclusion: nil,
-    on: nil,
-    unless: nil,
     if: nil,
-    allow_nil: nil
-  )
-  end
+    on: nil,
+    prepend: T::Boolean,
```
jaredbeck: In the `def`, should this be `prepend: nil`?

jaredbeck:
Validate can also take a block.

```ruby
class Comment
  include ActiveModel::Validations
  validate do |comment|
    comment.must_be_friends
  end
```

It looks like `sig` supports blocks:

```ruby
sig { params(blk: T.proc.returns(T::Boolean).void }
```

https://sorbet.org/docs/procs#annotating-block-parameters-with-yield

AlexEvanczuk:
thanks for the fixes!

lib/activerecord/all/activerecord.rbi
pje: Thank you!

pje:
```suggestion
    prepend: nil,
```

jaredbeck:
>    sig { params(blk: T.proc.returns(T::Boolean).void }

How do we specify an optional block? Use `T.nilable`?

pje:
> Validate can also take a block.

@jaredbeck Great catch! Thank you.

> How do we specify an optional block? Use T.nilable?

That works! You can also apparently just omit the blk arg from the sig? https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params(a%3A%20Integer).void%20%7D%0Adef%20foo(a%2C%20%26blk)%3B%20end%0A%0Afoo%201%20do%20%7Cx%2C%20y%2C%20z%7C%0A%20%20x%20%2B%20y%20%2B%20z%0Aend%0A

pje:
> sig { params(blk: T.proc.returns(T::Boolean).void }

@jaredbeck: looks like the block doesn't actually need to return a boolean. You can just use it for side-effects. Will make it `void` in that case.

pje:
Since we merged #55 (which included this original PR), this PR just adds correctly-typed default values for the optional keyword args.

I think this is ready to go!

PR #55, opened 2019-07-12
Move ActiveRecord::Base methods into their actual modules and include/extend them explicitly

pje:
### Why

Previously we'd just defined a lot of methods right inside `ActiveRecord::Base`—that worked initially but it's not correct or modular. If someone tried to include `ActiveModel::Validations::ClassMethods` directly, e.g., they wouldn't get our signatures correctly.

### How

This changeset just moves around those signatures into something more like the actual byzantine module hierarchy Rails uses.

### A Caveat on `Concerns`

Rails uses `Concerns` a lot, which is hard to model using Sorbet. When you `include` a `FooConcern`, `FooConcern` will try to automatically `extend`  your class with `FooConcern::ClassMethods`. Sorbet has [`mixes_in_class_methods`](https://sorbet.org/docs/abstract#interfaces-and-the-included-hook) to model modules that `extend` your class via their `included` hook, but I don't know of any way to model transitive `include`s (modules that include another module in their `included` hook). So I just explicitly listed those all as `includes` in `ActiveRecord::Base` and added comments to document that I hand-wrote them. It's not ideal!

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -191,6 +191,15 @@ module ActiveRecord::Associations::ClassMethods
   ); end
 end
 
+module ActiveRecord::AttributeMethods
+  include ActiveModel::AttributeMethods
+end
+
+module ActiveRecord::AttributeMethods::Serialization::ClassMethods
+  sig { params(attr_name: Symbol, class_name_or_coder: T.untyped).void }
+  def serialize(attr_name, class_name_or_coder = Object); end
+end
+
 module ActiveRecord::NestedAttributes::ClassMethods
   sig do
     params(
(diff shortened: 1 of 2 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -226,11 +235,68 @@ module ActiveRecord::Scoping::Named::ClassMethods
 end
 
 class ActiveRecord::Base
-  extend ActiveRecord::Associations::ClassMethods
-  extend ActiveRecord::Inheritance::ClassMethods
-  extend ActiveRecord::Scoping::Named::ClassMethods
-  extend ActiveRecord::NestedAttributes::ClassMethods
-  extend ActiveRecord::Persistence::ClassMethods
+  extend ActiveModel::Naming
```
pje: Initially, this list of `include`/`extend` modules was copy-pasted directly: https://github.com/rails/rails/blob/b9ca94cae/activerecord/lib/active_record/base.rb#L278-L325

Anything that didn't come from the rails source I added manually to model transitive concerns/includes/extends. (Those are all commented!)

aisamanra:
Thank you!

PR #60, opened 2019-07-13
Update ActiveRecord to remove 5.2 classes from all

AlexEvanczuk:
1) Added `ActiveRecord::Inheritance::ClassMethods` mixin so that we can use `self.abstract_class =`
2) Move 5.2 only classes out of `all`.


AlexEvanczuk:
pinging @aisamanra

PR #61, opened 2019-07-13
Added options to PaperTrail::Model::ClassMethods#has_paper_trail

nitsujri:
Fixes passing options to the `has_paper_trail` method.

Defined here: https://github.com/paper-trail-gem/paper_trail/blob/master/lib/paper_trail/has_paper_trail.rb#L66

### DIFF
--- a/lib/paper_trail/all/paper_trail.rbi
+++ b/lib/paper_trail/all/paper_trail.rbi
@@ -1,7 +1,7 @@
 # typed: strong
 
 module PaperTrail::Model::ClassMethods
-  def has_paper_trail
+  def has_paper_trail(options = {})
   end
 end
 


lib/paper_trail/all/paper_trail.rbi
```diff
@@ -1,7 +1,7 @@
 # typed: strong
 
 module PaperTrail::Model::ClassMethods
-  def has_paper_trail
+  def has_paper_trail(options)
```
aisamanra: This is going to make the `options` parameter obligatory, but it looks like here https://github.com/paper-trail-gem/paper_trail/blob/master/lib/paper_trail/has_paper_trail.rb#L66 it has a default value. If you change this to `options = {}`, then Sorbet will understand it as an optional positional argument: would the be more accurate?

nitsujri: @aisamanra you're absolutely right, but interestingly when I pass only 1 variable in, Sorbet doesn't complain.

1__justin_Justins

When I pass 2 in, of course it errors with `expected: 2, got: 0`, but I'm not sure why it doesn't complain with 1. If I had more time, I'd investigate.

Regardless, this has been updated to be actually optional.

DarkDimius:
Thank you!

PR #62, opened 2019-07-13
Update activemodel.rbi to allow format to be set to a Regexp.

connorshea:
`format` in `validates` actually also allows just a regex, e.g. I have code like this:

```ruby
validates :username,
  # Allow letters, numbers, disallow _ and . at the start or end,
  # disallow _ or . next to each other or themselves.
  format: /\A(?=.{4,20}\z)[a-zA-Z0-9]+(?:[._][a-zA-Z0-9]+)*\z/
```

And it works perfectly fine.


DarkDimius:
Thank you, Connor!

PR #63, opened 2019-07-14
Add some methods to activerecord.rbi that are used in migrations.

connorshea:
I noticed a lot of my migration files weren't typed simply because Sorbet didn't know about a lot of these methods, so I've fixed that.

https://edgeguides.rubyonrails.org/active_record_migrations.html

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -862,26 +862,39 @@ end
 
 class ActiveRecord::Migration::Current < ActiveRecord::Migration
   def create_table(table_name, options = nil); end
-  def drop_table(table_name, options = {}); end
   def change_table(table_name, options = {}); end
-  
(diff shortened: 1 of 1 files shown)


connorshea:
This is still missing a few methods, e.g. extensions, reversible, execute, and timestamps. I’ll add those when I get a chance tonight.

connorshea:
This should be good to go now 👍

connorshea:
Correction: I need to do a bit more work on this to make sure everything works. I'm missing a lot of options from the `*options` parameters in these.

connorshea:
On my Rails app, this change causes `bundle exec srb tc --typed=true --ignore spec/` to go from 240 errors to 109 🎉

DarkDimius:
> On my Rails app, this change causes bundle exec srb tc --typed=true --ignore spec/ to go from 240 errors to 109 🎉


wow, this is awesome! Thank you!

PR #38, opened 2019-06-28
Add more sigs for ActiveRecord

pje:
Adds basic sigs for:

```
ActiveRecord::Base.{has_one,has_many,belongs_to,has_and_belongs_to_many,scope}
ActiveRecord::Type::Value#cast
ActiveRecord::Type::Boolean#cast
```

Some of the sigs are kinda vague, but this is definitely progress!

PR #65, opened 2019-07-15
Fix small bug in type signature of `ActiveRecord::Type::Boolean#cast`

pje:
This was a copy-pasted error I added this in #38. The method stub shouldn't have a Sorbet type. :\


DarkDimius:
Thank you!

PR #68, opened 2019-07-17
Type Grape

AlexEvanczuk:
Basic types for https://github.com/ruby-grape/grape

Some discussion about this here:
https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1563314645490800

> Hi everyone, I’ve been discussing an interesting thing around typing gems and code in general with @Jeff Carbonella.  I am typing this gem in `sorbet-typed`: https://github.com/ruby-grape/grape.
> 
> What we’re finding is this gem uses a lot of fairly complex, somewhat hard to follow meta-programming techniques to get methods into the `Grape::API` class that all of our own application API code inherits from.
> 
> Originally, I was trying to figure out all of the implementation details of how the methods got into there, and tried to type the gem based on implementation details.
> 
> Later, we realized that we can simply type the methods in `Grape::API` and forget about the implementation details entirely.  We felt this was better because:
> A) Types should serve as documentation and hide implementation detail
> B) The inside implementation of a gem should be considered private API, and we should only type public API
> C) Public API types are kind of like an “integration test” over a “unit test.”  We don’t really need to unit test our code — we just want to make sure the available API works as expected (this is similar to B).
> 
> We welcome any thoughts on this and are curious if there are any opposing opinions on it! (edited)

### DIFF
--- a/lib/grape/all/grape.rbi
+++ b/lib/grape/all/grape.rbi
@@ -222,4 +222,3 @@ class Grape::API
   def self.format(format)
   end
 end
-
--- a/lib/grape/all/grape.rbi
+++ b/lib/grape/all/grape.rbi
@@ -1,0 +1,225 @@
+# typed: strong
+
+class Virtus::Attribute::Boolean
+end
+
+class Grape::API
+  Boolean = Virtus::Attribute::Boolean
+
+  sig do
+    params(
+      name: Symbol,
+      type: T.untyped,
+      desc: String,
+      allow_blank: T::Boolean,
+      values: T::Array[T.untyped],
+      date_parameter: T::Boolean,
+      positive_number: T::Boolean,
+      acord130_question_number: T::Boolean,
+      id_for_entity_type: T.untyped,
+      block: T.nilable(T.proc.void)
+    ).void
+  end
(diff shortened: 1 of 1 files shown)


DarkDimius:
Thank you!

lib/grape/all/grape.rbi
```diff
@@ -0,0 +1,225 @@
...
+
+  sig do
+    params(
+      format: Symbol,
+    ).void
+  end
+  def self.format(format)
+  end
+end
+
```
DarkDimius:
```suggestion
```

DarkDimius:
Thank you!

lib/grape/all/grape.rbi
```diff
@@ -0,0 +1,224 @@
...
+      )
+    ).void
+  end
+  def self.mount(api_class)
+  end
+
+  sig do
+    params(
+      object: T.untyped,
+      with: T.nilable(T.class_of(Grape::Entity))
```
ghiculescu, line 190: i think this line will crash if you don't have the https://github.com/ruby-grape/grape-entity gem installed

PR #69, opened 2019-07-17
Improve ActiveRecord::Persistence method signatures

pje:
Improves some of the signatures for the `ActiveRecord::Persistence` module.

- `ActiveRecord::Persistence#becomes` should return `T.untyped`, not `Class`, since it'll try to return an instance of the specified class
- `ActiveRecord::Persistence#increment` actually returns `self`
- `ActiveRecord::Persistence#toggle` actually returns `self`
- `ActiveRecord::Persistence#toggle!` actually only returns `true` (or raises an error)
- `ActiveRecord::Persistence#update!` actually only returns `true` (or raises an error)
- None of the methods defined in `ClassMethods` can return `T.self_type`, because that'll actually refer to the class, not an instance. We can't generically specify that they should return an _instance_ of `T.self_type`, so they've gotta be T.untyped instead until that feature's implemented in Sorbet.
- The methods `becomes`, `becomes!`, `decrement`, `decrement!`, `increment`, `increment!`, `new_record?`, `persisted?`, `reload`, `save!`, `save`, `toggle!`, `toggle`, `touch`, `update_attribute`, `update_column`, `update_columns`, and `update` are instance methods defined in the `Persistence` module; they should never have been defined in `ClassMethods`. That was a bug!

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -522,36 +522,12 @@ end
 
 module ActiveRecord::Persistence
   mixes_in_class_methods(ActiveRecord::Persistence::ClassMethods)
-end
 
-module ActiveRecord::Persistence::ClassMethods
-  sig { params(klass: Class).returns(Class) }
+  sig { params(klass: Class).returns(T.untyped) }
   def becomes!(klass); end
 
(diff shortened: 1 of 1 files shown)


frewsxcv:
this was a Nice change, thanks patrick!

PR #70, opened 2019-07-18
Removed trailing blanks on 2 lines in bundler.rbi file

jasnow:
Removed trailing blanks on 2 lines in bundler.rbi file so overcommit stops complaining.


aisamanra:
I'm going to hold off on merging this for the moment because, due to some current weaknesses that we're addressing in our CI system, this change actually has the potential to cause temporary test breakages. (We diff line-by-line against newly-initialized Sorbet projects, and those will include the Bundler RBI, so changes to the Bundler RBI will cause problems with that test.) I'll have to merge it simultaneously with other changes, which I'll get around to soon!

jasnow:
Thanks

PR #71, opened 2019-07-19
No bare generics

jez:


pje:
@jez: I thought bare generics were supposed to be equivalent—is that a change in Sorbet, or is this PR a style thing?

pje:
I've just been linked to the corresponding Sorbet PR: https://github.com/sorbet/sorbet/pull/1331

aisamanra:
@pje There's a bit more context in my comment here https://github.com/sorbet/sorbet/pull/1309#issuecomment-512616370, if you want to get a bit of the history as well.

PR #73, opened 2019-07-22
Add more versions to ActiveRecord.

connorshea:
Renamed ~>5.2 to >=5.2, since 6.0 also has these includes/extends.

And then I split the ActiveRecord version aliases between version-specific files.

### DIFF
--- a/lib/activerecord/>=5.2/activerecord.rbi
+++ b/lib/activerecord/>=5.2/activerecord.rbi
@@ -1,7 +1,5 @@
 # typed: strong
 
-ActiveRecord::Migration::Compatibility::V5_2 = ActiveRecord::Migration::Current
-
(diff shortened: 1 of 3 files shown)


connorshea:
I just noticed `>=` isn't mentioned in the README, is it actually supported?

jez:
> I just noticed `>=` isn't mentioned in the README, is it actually supported?

Yeah, we just take what's in the folder name and hand it directly to `Gem::Requirement`. From the README:

> The naming of this directory follows the format of `Gem::Requirement`.

And from the ruby docs:

> A Requirement is a set of one or more version restrictions. It supports a few (=, !=, >, <, >=, <=, ~>) different restriction operators.

https://docs.ruby-lang.org/en/2.6.0/Gem/Requirement.html

PR #74, opened 2019-07-22
Add ActiveSupport's String and Array methods

connorshea:
Also format the signatures consistently.

Based on the ActiveSupport String docs: https://api.rubyonrails.org/classes/String.html

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -1,7 +1,7 @@
 # typed: strong
 
 module ActiveSupport
-  sig {params(kind: Symbol, blk: T.proc.bind(T.class_of(ActionController::Base)).void).void}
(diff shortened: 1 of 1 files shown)


lib/activesupport/all/activesupport.rbi
```diff
@@ -96,5 +96,215 @@ class NilClass
...
   end
-  def try!(method_name = nil, *args, &b) ; end
+  def try!(method_name = nil, *args, &b); end
+end
+
+class String
+  sig { returns(T::Boolean) }
+  def acts_like_string?; end
+
+  sig { params(position: T.any(Integer, String, Regexp)).returns(T.nilable(String)) }
```
elliottt: Why not support the `Range` case as well, for the position argument?

connorshea: Good catch, I missed that Ranges were accepted. Fixed 👍

PR #76, opened 2019-07-22
Fix typo in `validate` method stub

pje:
This should've been `false` the whole time—that was a bug in my initial commit.


DarkDimius:
Thank you!

PR #78, opened 2019-07-22
Add type signatures for ActiveSupport::NumberHelper.

connorshea:
Based on https://api.rubyonrails.org/classes/ActiveSupport/NumberHelper.html


DarkDimius:
Thank you!

PR #82, opened 2019-07-24
 Removed trailing blanks on 1 line in activesupport.rbi file

jasnow:
Removed trailing blanks on 1 line in activesupport.rbi file to avoid overcommit trigger.


jasnow:
Thanks

PR #86, opened 2019-07-26
Tighten some ActiveRecord sigs

pje:
This simplifies some of the ActiveRecord sigs, corrects a few small method parameter omissions, and refactors a bit.

- Optional keyword args shouldn't be `nilable` unless `nil` is actually an expected value people can pass in. With few exceptions, it wouldn't actually make sense for people to explicitly pass in `nil` as a value for these keyword args. So I simplified a lot of these sigs by removing the `nilable` from most keyword args
- All callbacks (after_create, after_save, etc..) accept an optional block
- All callbacks (after_create, after_save, etc..) allow the required positional arg to be either a Symbol or a Proc. Same goes for the optional :if and :unless keyword args
- All callbacks (after_create, after_save, etc..) allow the required positional arg to be a symbol or a proc
- Add an UntypedCallbackProc type alias, just to make the sigs a little more readable

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -18,32 +18,34 @@ VariadicUntypedFunction = T.type_alias(
 )
 
 module ActiveRecord::Associations::ClassMethods
+  extend T::Sig
+
   sig do
     params(
       name: Symbol,
-      scope: T.nilable(T.proc.void),
-      after_add: T.nilable(T.any(Symbol, String, T.proc.void)),
-      after_remove: T.nilable(T.any(Symbol, String, T.proc.void)),
-      anonymous_class: T.nilable(T.any(Symbol, String)),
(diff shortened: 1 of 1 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -701,34 +741,34 @@ module ActiveRecord::Persistence::ClassMethods
   sig do
     params(
       attributes: T::Array[T::Hash[T.any(Symbol, String), T.untyped]],
-      returning: T.nilable(T.any(FalseClass, T::Array[T.any(Symbol, String)]))
```
elliottt: The docs show that this has a `nil` default argument, why remove that information from the signature?

connorshea:
FYI, #87 caused merge conflicts

DarkDimius:
Closing this as stale, please feel free to reopen.

PR #88, opened 2019-07-28
Add types for activerecord-safer_migrations

clarkdave:
Add types for `activerecord-safer_migrations` gem

- https://github.com/gocardless/activerecord-safer_migrations
- https://rubygems.org/gems/activerecord-safer_migrations

These are compatible with both v2 (current) and v1.


DarkDimius:
Thank you, Dave

PR #90, opened 2019-07-28
Rename ~>6.0.0 to ~>6.0.0.rc1 for activerecord.

connorshea:
Previously, release candidate versions of Rails 6 weren't included with this range, so new migrations created on Rails 6 would fail typechecking.

This is a quirk with how RubyGems does version ranges involving prereleases. I've confirmed that this will continue to work when 6.0.0 final comes out (if you try to require Rails `~>5.2.0.rc1` in a Gemfile it'll install Rails 5.2.3).


DarkDimius:
Thank you

PR #91, opened 2019-07-29
Create minitest.rbi

frewsxcv:
https://github.com/seattlerb/minitest

### DIFF
--- a/lib/minitest/all/minitest.rbi
+++ b/lib/minitest/all/minitest.rbi
@@ -1,0 +1,93 @@
+# typed: strong
+
+module Minitest
+  class Test < Runnable
+    include Minitest::Assertions
+    include Minitest::Reportable
+  end
+end
+
+module Minitest::Assertions
+  extend T::Sig
+
+  sig { params(test: T.untyped, msg: T.nilable(String)).returns(TrueClass) }
+  def assert(test, msg = nil); end
+
+  sig do
+    params(
+      exp: Object,
+      msg: T.nilable(String)
+    ).returns(TrueClass)
+  end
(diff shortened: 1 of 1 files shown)


lib/minitest/all/minitest.rbi
```diff
@@ -0,0 +1,93 @@
+# typed: strong
+
+module Minitest
+  class Test < Runnable
+    include Minitest::Assertions
+    include Minitest::Reportable
```
elliottt: Where is `Minitest::Reportable` defined?

lib/minitest/all/minitest.rbi
```diff
@@ -0,0 +1,93 @@
...
+
+module Minitest::Assertions
+  extend T::Sig
+
+  sig { params(test: T.untyped, msg: T.nilable(String)).returns(TrueClass) }
+  def assert(test, msg = nil); end
+
+  sig do
+    params(
+      exp: Object,
```
jez: It's possible you want to use `BasicObject` instead of `Object` in all of these sigs, if you want to say "you can pass literally anything here" (`class Object < BasicObject` in Ruby)

frewsxcv: sounds right! addressed in https://github.com/sorbet/sorbet-typed/pull/91/commits/cdda2a566822e10c511db77dade8cecb3cbee910

lib/minitest/all/minitest.rbi
frewsxcv: hmm good point. removed in https://github.com/sorbet/sorbet-typed/pull/91/commits/1e212927b6b0f125057ede4f5a101973406f41d5

PR #92, opened 2019-08-02
Add a basic gem.rbi to lib/ruby.

connorshea:
The goal here is to resolve https://github.com/sorbet/sorbet/issues/1455

There are a few problems with this as it is now:

- Almost everything is untyped
- Lots of `_` parameters
- This is specifically for RubyGems 3.x
- I'm not sure if `gem.rbi` should be part of the `ruby/` directory or not?
- I'm not sure if `gem.rbi` should actually be `rubygems.rbi`?
- Can Sorbet actually determine the user's RubyGems version?
- There may be some minor errors, I generated this with Sord, but I've fixed a few of its quirks manually and generally I think it's been pretty well battle-tested at this point.

Adding this RBI to my `sorbet/` directory and then regenerating `hidden.rbi` drops it down from 7419 lines to 3386 lines.

### DIFF
--- a/lib/ruby/all/gem.rbi
+++ b/lib/ruby/all/gem.rbi
@@ -6,473 +6,320 @@ module Gem
   RubyGemsVersion = T.let(nil, T.untyped)
   USE_BUNDLER_FOR_GEMDEPS = T.let(nil, T.untyped)
 
-  sig { params(path: T.untyped).returns(T.untyped) }
   def self._deprecated_detect_gemdeps(path = T.unsafe(nil)); end
 
-  sig { params(data: T.untyped).returns(T.untyped) }
   def self._deprecated_gunzip(data); end
 
-  sig { params(data: T.untyped).returns(T.untyped) }
   def self._deprecated_gzip(data); end
 
-  sig { params(data: T.untyped).returns(T.untyped) }
   def self._deprecated_inflate(data); end
 
-  sig { params(name: T.untyped, args: T.untyped).returns(T.untyped) }
   def self.activate_bin_path(name, *args); end
 
-  sig { params(base_dir: T.untyped).returns(T.untyped) }
   def self.default_ext_dir_for(base_dir); end
 
-  sig { returns(T::Boolean) }
   def self.default_gems_use_full_paths?; end
 
-  sig { returns(T.untyped) }
   def self.default_spec_cache_dir; end
 
-  sig { params(data: T.untyped).returns(T.untyped) }
   def self.deflate(data); end
 
-  sig { params(args: T.untyped, block: T.untyped).returns(T.untyped) }
   def self.detect_gemdeps(*args, &block); end
 
-  sig { returns(T.untyped) }
   def self.dir; end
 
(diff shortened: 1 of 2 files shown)


connorshea:
Also, this exists already: https://github.com/sorbet/sorbet/blob/719eb639e535860a3f1e3288fed3dcd47586eb5e/rbi/stdlib/gem.rbi

So that's probably a problem

connorshea:
Any ideas on fixing this last error? 🤔 
```
Connors-MacBook-Pro:sorbet-typed connorshea$ ruby .ci/run.rb 
lib/bundler/all/bundler.rbi:1211: Type Elem declared by parent T.class_of(Gem::Specification) must be re-declared in T.class_of(Bundler::EndpointSpecification) https://srb.help/5036
    1211 |class Bundler::EndpointSpecification < Gem::Specification
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    https://github.com/sorbet/sorbet/tree/719eb639e535860a3f1e3288fed3dcd47586eb5e/rbi/stdlib/gem.rbi#L207: Elem declared in parent here
     207 |class Gem::Specification < Gem::BasicSpecification
                ^^^^^^^^^^^^^^^^^^
Errors: 1
```

jasnow:
Did you try "srb tc--suggest-typed" or "srb tc -a" ?

jez:
@connorshea I would propose removing the autogenerated sigs that have untyped. Putting them there locks in the sigs for everyone else, whereas by omitting them, people can one-off sig individual methods to have better types in the mean time while they're submitting a patch to add proper types upstream.

connorshea:
Removed all the untyped stuff 👍

elliottt:
It looks like the error with the bundler rbi is due to the bundler.rbi that exists in sorbet https://github.com/sorbet/sorbet/blob/master/rbi/gems/bundler.rbi

connorshea:
Hmm, why do you think so?

elliottt:
The build failure for your most recent commit says that you need to declare `Elem`, but you're already doing that as of a86b6ec2379d5cbc6b880cddc70c30f822a21529.

connorshea:
Any ideas on how to fix it?

lib/ruby/all/gem.rbi
```diff
@@ -0,0 +1,4568 @@
...
+    def suggestion; end
+
+    def suggestion=(suggestion); end
+  end
+
+  module Security
+    AlmostNoSecurity = T.let(nil, T.untyped)
+    DIGEST_NAME = T.let(nil, T.untyped)
+    EXTENSIONS = T.let(nil, T.untyped)
+    H
```
elliottt: Did you mean to `extend Enumerable` here? That's likely what's causing problems with the bundler rbi.

connorshea: oh wow, good catch. I'm not entirely sure how that got in there.

elliottt: You should also be able to remove the changes to the bundler rbi now.

connorshea: 👍 done

PR #93, opened 2019-08-02
Add #request method for Rails controllers

panthomakos:
This is a first step in filling out the ActionController sigs for Rails. Technically the `#request` method is nillable. In practice I don't think this is intended to be the case (at least the docs indicate that a request object is returned). That said, it's entirely possible to instantiate a controller with a `nil` request outside of a normal Rails request:

```
::AplicationController.new.request # => nil
```

This results in the need to use `T.nilable(ActionDispatch::Request)` in type signatures or `T.must` at the call-site to avoid nils. Some guidance on if I can omit the `T.nilable` would be much appreciated.

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -1,5 +1,16 @@
 # typed: strong
 
+class ActionController::Base < ActionController::Metal
+end
+
+class ActionController::Metal
+  sig {returns(ActionDispatch::Request)}
+  def request(); end
+end
+
+class ActionDispatch::Request
+end
+
 class ActionDispatch::Routing::RouteSet
   sig {params(blk: T.proc.bind(ActionDispatch::Routing::Mapper).void).void}
   def draw(&blk); end


connorshea:
I tried this on my Rails app, and I definitely think it shouldn't be marked as nilable. I feel like `request` would so rarely be nilable that it's not worth the trouble of adding `T.must` everywhere.

panthomakos:
Thanks! Made the adjustments to that the return type is not nilable. Please take another look.

connorshea:
LGTM 👍

connorshea:
Though actually, aren't `request` and `response` methods of `ActionController::Base` rather than `ActionController::Metal`? https://api.rubyonrails.org/classes/ActionController/Base.html#method-i-request

panthomakos:
They are defined on `Metal`: https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal.rb

`Base` inherits them: https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/base.rb

PR #100, opened 2019-08-06
Add Hash extensions to ActiveSupport

frewsxcv:


manhhung741:
I think the return type T.self_type on those methods are incorrect.

lib/activesupport/all/activesupport.rbi
```diff
@@ -417,3 +417,35 @@ module ActiveSupport::NumberHelper
...
+  sig { returns(T.self_type) }
+  def deep_symbolize_keys!; end
+
+  sig { returns(T.self_type) }
+  def deep_transform_keys; end
+
+  sig { returns(T.self_type) }
+  def deep_transform_keys!; end
+
+  sig { returns(T.self_type) }
```
manhhung741: This should change the key of a Hash from any -> string. `T.self_type` is wrong

lib/activesupport/all/activesupport.rbi
```diff
@@ -417,3 +417,35 @@ module ActiveSupport::NumberHelper
...
+  def stringify_keys; end
+
+  sig { returns(T.self_type) }
+  def stringify_keys!; end
+
+  sig { returns(T.self_type) }
```
manhhung741: This should change the key of a hash from any -> symbol. `T.self_type` is incorrect

lib/activesupport/all/activesupport.rbi
frewsxcv: ah you're right. are you able to open a pr?

manhhung741:
I haven’t found a good solution yet. Ideally if this is on a T::Hash we can
change the type of the Key and keep the type of the value. But this is on a
Hash.

My temporary patch is to remove these methods.

frewsxcv:
> My temporary patch is to remove these methods.

Why would we remove these methods? Why wouldn't we just use `T.untyped`?

manhhung741:
It would be the same thing. It’s for my local copy so it doesnt really
matter.

manhhung741:
So this actually works. I'll send out a PR later!


class Hash < Object
extend T::Sig

sig { returns(T::Hash[Symbol, V]) }
def symbolize_keys
T.unsafe(self)
end
end

ghiculescu:
@manhhung741 what results did you get there? Maybe I'm misunderstanding, but I was hoping that in this example, the type of the value would be retained, rather than reverting to `T.untyped`.

```ruby
# hash.rbi

# typed: strict

class Hash
  extend T::Sig
  sig { returns(T::Hash[String, V]) }
  def stringify_keys; end
end
```

```ruby
# test.rb

# typed: true
require 'active_support/all'
require 'sorbet-runtime'

extend T::Sig

sig { returns( {a: String} ) }
def get_specific_hash_shape
  {a: "1"}
end

T.reveal_type(get_specific_hash_shape) # Revealed type: {a: String}
T.reveal_type(get_specific_hash_shape.stringify_keys) # T::Hash[String, T.untyped]

```

ghiculescu:
Also @manhhung741 FYI on https://github.com/sorbet/sorbet-typed/pull/148

PR #101, opened 2019-08-06
Extend ActionController::Parameters types

frewsxcv:


lib/actionpack/all/actionpack.rbi
```diff
@@ -43,12 +43,57 @@ class ActionController::Parameters
...
+    ).returns(
+      T.any(
+        String,
+        Numeric,
+        T::Array[T.untyped],
+        ActionController::Parameters,
+      ),
+    )
+  end
+  def require(key); end
```
connorshea: This redefines `require` again, instead of `required`.

lib/actionpack/all/actionpack.rbi
```diff
@@ -43,12 +43,57 @@ class ActionController::Parameters
...
+    params(
+      key: T.any(
+        String,
+        Symbol,
+        T::Array[T.any(String, Symbol)],
+      ),
+    ).returns(
+      T.any(
+        String,
+        Numeric,
```
connorshea: Does it actually return Strings/Numbers? This makes it much harder to write code like so:

```ruby
def book_params
  params.require(:book).permit(:author_id, :title)
end
```

Since `permit` won't exist on the other types this can return.

lib/actionpack/all/actionpack.rbi
frewsxcv: good catch! fixed in https://github.com/sorbet/sorbet-typed/pull/101/commits/118b8ddbd29dd5a894ee8aa829478c2f497f0e5b

lib/actionpack/all/actionpack.rbi
frewsxcv: > Does it actually return Strings/Numbers?

yep

```ruby
pry(main)> ActionController::Parameters.new({foo: 1}).require(:foo)
=> 1

pry(main)> ActionController::Parameters.new({foo: 'foo'}).require(:foo)
=> "foo"

pry(main)> ActionController::Parameters.new({foo: {bar: 'baz'}}).require(:foo)
=> <ActionController::Parameters {"bar"=>"baz"} permitted: false>

pry(main)> ActionController::Parameters.new({foo: 1, bar: 1}).require([:foo, :bar])
=> [1, 1]
```

> This makes it much harder to write code like so:

oh totally agreed. i wrote these helpers a week ago which have helped with the types: https://gist.github.com/frewsxcv/fc5d7426921bb64f5a6821721cff4581

connorshea: that's... really weird.

How does `params.require(:foo).permit(:bar)` in controllers work, then? Is `params` actually something other than an instance of `ActionController::Parameters`?

connorshea: Oh, it seems to looks like this:

```ruby
ActionController::Parameters.new({foo: { bar: 'test'}}).require(:foo).permit(:bar)
=> <ActionController::Parameters {"bar"=>"test"} permitted: true>
```

DarkDimius:
@frewsxcv Hi Correy,
 thank you for your contribution.
There are some merge conflicts, can you help resolve them?

frewsxcv:
@DarkDimius done!

lib/actionpack/all/actionpack.rbi
frewsxcv: yeah, this is a gnarly api. i don't really understand how to write fullproof code with it, without a bunch of `is_a?` checks...

frewsxcv:
oh hey! test failures! fixing...

lib/actionpack/all/actionpack.rbi
frewsxcv: also, adding onto this, since `params` can [be populated from a json request](https://guides.rubyonrails.org/action_controller_overview.html#json-parameters), any scalar json datatype can be returned from `require`, including a number

DarkDimius:
@connorshea , as this modifies the test you've introduced, could you confirm if changes are fine?

Timing out in 2 business days.

connorshea:
@DarkDimius part of me wants to block this so I don't have to deal with it :P

But really, it's a more accurate signature so I'd say it should probably be merged.

lib/actionpack/all/actionpack.rbi
```diff
@@ -132,11 +132,56 @@ class ActionController::Parameters
   sig { params(block: T.untyped).returns(T.untyped) }
   def reject(&block); end
 
-  sig { params(key: T.any(Symbol, T::Array[Symbol])).returns(ActionController::Parameters) }
+  sig do
```
manhhung741: This is more accurate. Returning `ActionController::Parameters` is outright wrong and can make people distrust the type system.

manhhung741:
I feel like some of the return type should just be `T.untyped`. That might be more useful that having a bunch of return types that is really hard to handle.

Am I expected to like handle all different types of `params.require(:id)`? I don't think it's useful. It's probably more useful for people to `T.let` or `T.cast` the result to a type so that it can be asserted at runtime.

frewsxcv:
Merge conflicts addressed!

Issue #42, opened 2019-07-03
Set up spec for the rbi


manhhung741:
There should be a way to test the rbi so that reviewing PR is easier.
I expect it would be:
- Set up sorbet in this repo
- Set up structure to add testcase & output from `srb tc`

When people add rbi, they should also add test cases & expected output

connorshea:
I was curious how DefinitelyTyped does this, and it looks like they have a tests file for every package that includes some basic code for the package, to make sure it passes static type checking.

e.g. [`actioncable/index.d.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d0c0725ae4b6e0306204f37e291e90f36daca14c/types/actioncable/index.d.ts) and [`actioncable-tests.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d0c0725ae4b6e0306204f37e291e90f36daca14c/types/actioncable/actioncable-tests.ts)

That seems like the simplest solution to me.

jasnow:
Can we run "lint" on the rbi files as a first step at testing?

PR #102, opened 2019-08-06
Allow `path => controller#action` in Routes

panthomakos:
The `'path' => 'controller#action'` is a supported short-hand syntax
for Rails routes as documented [here][1]. The current type is too
restrictive and does not support this valid syntax. This change modifies
the first argument type so that it supports the Hash syntax.

[1]: https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/routing/mapper.rb#L1605

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -111,7 +111,7 @@ module ActionDispatch::Routing::Mapper::HttpHelpers
   # ActionDispatch::Routing::Mapper::Resources#match
   sig do
     params(
-      name: T.any(String, Symbol),
+      name: T.any(String, Symbol, T::Hash[String, String]),
       controller: T.nilable(T.any(String, Symbol)),
       action: T.nilable(T.any(String, Symbol)),
       param: T.nilable(Symbol),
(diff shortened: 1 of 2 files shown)


panthomakos:
Might be worth considering adding a test-suite for some of these routes. For example, a simple `routes.rb` file with some variations that could be type-checked.

connorshea:
Yup, that's what #42 is for. Though I suppose we could still have files that use the library and make sure all valid input is allowed 🤔 That shouldn't be as hard as porting the test harness over to this repo.

panthomakos:
I rebased on `master` and included tests that fail before the change:

```
lib/actionpack/test/actionpack_test.rb:40: Expected T.any(String, Symbol) but found {String("about") => String("static_pages#about")} for argument name https://srb.help/7002
    40 |  get 'about' => 'static_pages#about'
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    lib/actionpack/all/actionpack.rbi:151: Method ActionDispatch::Routing::Mapper::HttpHelpers#get has specified name as T.any(String, Symbol)
     151 |      name: T.any(String, Symbol),
                ^^^^
  Got {String("about") => String("static_pages#about")} originating from:
    lib/actionpack/test/actionpack_test.rb:40:
    40 |  get 'about' => 'static_pages#about'
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

PR #95, opened 2019-08-04
Add many more methods and sigs for ActionPack

connorshea:
I'm leaving this as a draft until #93 is merged. It builds off that PR, and I'll rebase once the PR is merged.

I used these sigs with my Rails app and it seems like there aren't any problems, but if anyone else has a Rails app to test it on, that'd be appreciated :)

Most of these changes are focused on improving the signatures for the `params` method used in controllers and methods used in `config/routes.rb`, as well as a few other controller-related methods.

The main classes/modules that I've added methods for are:

- https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/HttpHelpers.html
- https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Resources.html
- https://api.rubyonrails.org/classes/ActionController/StrongParameters.html
- https://api.rubyonrails.org/classes/ActionController/Parameters.html
- https://api.rubyonrails.org/classes/ActionController/Flash/ClassMethods.html

And here's a simplified example of the `routes.rb` file in my Rails app:
```ruby
# typed: strict
# For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
Rails.application.routes.draw do
  root 'home#index'
  get 'home/index'

  resources :games do
    get :search, on: :collection
    post :add_game_to_library, on: :member
    delete :remove_game_from_library, on: :member
    delete :remove_cover, on: :member
    post :favorite, on: :member
    delete :unfavorite, on: :member
  end

  resources :game_purchases, except: [:edit, :new] do
    post :bulk_update, on: :collection
  end

  resources :users do
    get :index, on: :collection
    get :show, on: :member
    get :statistics, on: :member
    post :update_role, on: :member
    delete :remove_avatar, on: :member
    post :connect_steam, on: :member
    delete :disconnect_steam, on: :member
    post :steam_import, on: :member
    delete :reset_game_library, on: :member
    get '/compare/:user_id...:other_user_id', as: :compare, action: :compare, on: :collection
  end

  resources :platforms do
    get :search, on: :collection
  end

  namespace :search do
    get :index, as: '/', path: '/'
  end

  namespace :settings do
    get :profile, as: '/', path: '/'
    get :account
    get :connections
  end

  get '/about', to: 'static_pages#about'
end
```

PR #104, opened 2019-08-06
Add test files for actionpack, activemodel, activerecord, and activesupport

connorshea:
I have no idea what this does to sorbet if this is merged and you try to update sorbet-typed.

I've included an intentionally failing method to show that this is working. However, it only allows files that have no type errors, there's no way to test that sorbet catches incorrect code.

This is a simplified version of my `routes.rb` file that I showed in #95.

The idea to do this was started by #42 and re-upped in #102 by @panthomakos.

cc: @DarkDimius @jez I know this isn't the test harness you wanted, and I'd still like to get that eventually, but I'm curious if you think this solution would be workable? Do you see any significant flaws with this approach?

### DIFF
--- a/.ci/run.rb
+++ b/.ci/run.rb
@@ -3,7 +3,7 @@
 # Exclude files if necessary here.
 exclude = []
 
-file_paths = Dir.glob("**/*.rbi").select{|p| !exclude.include?(p)}.map{|p| "\"#{p}\""}.join(' ')
+file_paths = Dir.glob("**/*.rb{i,}").select{|p| !exclude.include?(p)}.map{|p| "\"#{p}\""}.join(' ')
 
 srb_cmd = 'tc'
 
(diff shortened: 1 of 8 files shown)


connorshea:
I changed the git remote of sorbet-typed in my `/.cache/sorbet/sorbet-typed` directory, and it does not pull in anything from `test/` as far as I can tell

connorshea:
Alright, I've added more tests and explained how the tests work in the README. Plus I've fixed a few bugs I found with the existing signatures :) This should be good to review now.

connorshea:
@panthomakos 👍 you should open a PR to change that, it looks like a good idea (though I'd prefer if the files were named `gem_name_test.rb` still, and you'll have to check whether Sorbet pulls down the files properly when you organize the project like that.

You can go into `.cache/sorbet/sorbet-typed/` and use `git remote set-url origin URL_GOES_HERE` to change the git remote that Sorbet uses to get sorbet-typed stuff, and you'll also need to merge the change into the master branch on your fork since sorbet-typed can't be configured to use a different branch. Then update the sorbet-typed files in a project and you'll get whatever is in your fork of the project.

I really should open an issue to make this easier...

panthomakos:
Thank you for the write-up. I'll put a PR together!

PR #108, opened 2019-08-09
ActiveRecord update/update_attributes methods accept ActionController::Parameters

connorshea:
It's somewhat common to have code like this in a Rails controller:
```ruby
class GenresController < ApplicationController
  def update
    @genre = Genre.find(params[:id])
    authorize @genre

    if @genre.update(genre_params)
      redirect_to @genre, success: "#{@genre.name} was successfully updated."
    else
      flash.now[:error] = "Unable to update genre."
      render :edit
    end
  end
  
  private

  def genre_params
    params.require(:genre).permit(
      :name,
      :description
    )
  end
end
```

`genre_params` returns an instance of ActionController::Parameters, so if I want to add a sig to the method that says that, `@genre.update(genre_params)` would have a type error. I use this quite a lot in my Rails app, so I know it works.

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -638,7 +638,10 @@ module ActiveRecord::Persistence
 
   sig do
     params(
-      attributes: T::Hash[T.any(Symbol, String), T.untyped]
+      attributes: T.any(
+        T::Hash[T.any(Symbol, String), T.untyped],
+        ActionController::Parameters
+      )
     ).returns(TrueClass)
   end
   def update!(attributes); end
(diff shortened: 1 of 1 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -638,30 +638,42 @@ module ActiveRecord::Persistence
 
   sig do
     params(
-      attributes: T::Hash[T.any(Symbol, String), T.untyped]
+      attributes: T.any(
+        T::Hash[T.any(Symbol, String), T.untyped],
+        ActionController::Parameters
```
elliottt: It doesn't look like `ActionController::Parameters` has a sig for `[]` on it currently, would you like to add that one as well?

connorshea: Added a sig for it 👍

connorshea:
I've removed the `[]` method sig since #111 covers it and I want to avoid conflicts if I can.

PR #109, opened 2019-08-09
Run tests for each gem + version in isolation

panthomakos:
As the test harness is currently setup, if one version of a gem has a
signature that conflicts with another version of the same gem (or any
other gem), then the tests will never succeed because the two versions
will conflict with each other. This change resolves these issues by
running each gem and version combination (along with the respective
`all` directory for that gem) in isolation. A failure to type-check any
one gem + version combination will fail the test harness script.

### DIFF
--- a/lib/activemodel/all/activemodel_test.rb
+++ b/lib/activemodel/all/activemodel_test.rb
@@ -17,7 +17,7 @@ module ActiveModelTest
 
   validates :name, :login, :email, presence: true
   validates :order, presence: true
-  
+
   validates :name, :login, :email, absence: true
 
   validates :email, uniqueness: true
@@ -37,7 +37,7 @@ module ActiveModelTest
   validates :age, numericality: true, on: :update
 
   validates :token, presence: true, uniqueness: true, strict: TokenGenerationException
-  
+
   validates :card_number, presence: true, if: :paid_with_card?
 
   # TODO: These are valid but currently fail typechecking.
--- a/lib/activerecord/all/activerecord_test.rb
+++ b/lib/activerecord/all/activerecord_test.rb
@@ -8,7 +8,7 @@ class ActiveRecordMigrationsTest < ActiveRecord::Migration::Current
     create_table :products do |t|
       t.string :name
       t.text :description
- 
+
       t.timestamps
     end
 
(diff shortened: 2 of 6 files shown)


lib/activemodel/all/activemodel_test.rb
```diff
@@ -17,7 +17,7 @@ module ActiveModelTest
 
   validates :name, :login, :email, presence: true
   validates :order, presence: true
-  
+
```
panthomakos: My editor removes extra whitespace. I can revert these changes if preferred.

lib/activerecord/all/activerecord_test.rb
```diff
@@ -61,14 +61,13 @@ def change
...
+
     say "Created a table"
- 
+
     suppress_messages { add_index :products, :name }
     say "and an index!", true
- 
+
     say_with_time 'Waiting for a while' do
-      sleep 10
```
panthomakos: This was raising a type error when run in isolation:

```
lib/activerecord/all/activerecord_test.rb:71: Method sleep does not exist on ActiveRecordMigrationsTest https://srb.help/7003
    71 |      sleep 10
              ^^^^^^^^
    https://github.com/sorbet/sorbet/tree/80e1b24dadafc4ead575cb5e3166a691d2eb73e7/rbi/core/kernel.rbi#L800: Did you mean: Kernel#sleep?
     800 |  def sleep(duration); end
            ^^^^^^^^^^^^^^^^^^^
Errors: 1
```

I suspect that this is because we are no longer running with other gems at the same time. I viewed this as a positive result because the same issue could occur in RBI files and inadvertently cause a type error.

connorshea:
Oh, you should probably update the README

panthomakos:
Updated the README.

DarkDimius:
@panthomakos unfortunately, there are some conflicts now. Can you take a look?

lib/activerecord/all/activerecord_test.rb
jasnow: @panthomakos - Would "Kernel.sleep 10" work?

panthomakos:
@DarkDimius conflicts resolved - changed rebased.

lib/activerecord/all/activerecord_test.rb
panthomakos: Yes, but the call is not run and does not impact the return value of the function so it's not necessary in the first place.

DarkDimius:
Thank you!

PR #111, opened 2019-08-10
ActionController::Parameters methods

connorshea:
Methods defined here: https://api.rubyonrails.org/classes/ActionController/Parameters.html

I tested it with my Rails app to verify that it didn't break anything.

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -42,29 +42,14 @@ class ActionController::Parameters
   sig { params(key: T.any(String, Symbol), value: T.untyped).void }
   def []=(key, value); end
 
-  sig { returns(T.untyped) }
+  sig { returns(T.nilable(T::Array[T.any(String, Symbol)]))) }
   def always_permitted_parameters; end
 
-  sig { params(obj: T.untyped).returns(T.untyped) }
+  sig { params(obj: T.nilable(T::Array[T.any(String, Symbol)])).void }
   def always_permitted_parameters=(obj); end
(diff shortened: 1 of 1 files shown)


DarkDimius:
Thank you!

lib/actionpack/all/actionpack.rbi
```diff
@@ -33,35 +33,203 @@ class ActionDispatch::Response
...
+  sig { params(obj: T::Boolean).void }
+  def self.permit_all_parameters=(obj); end
+
+  sig { params(keys: T.any(String, Symbol)).returns(ActionController::Parameters) }
+  def slice!(*keys); end
+
+  sig { params(keys: T.any(String, Symbol)).returns(ActionController::Parameters) }
+  def slice(*keys); end
+
+  sig { returns(T.nilable(ActiveSupport::HashWithIndifferentAccess)) }
```
manhhung741: @connorshea Do you know when this will return `nil`? Coz making them returning non-nilable object will be very useful for me.
```
=> <ActionController::Parameters {"a"=>1, "b"=>2} permitted: false>
irb(main):006:0> param.permit().to_hash
=> {}

irb(main):007:0> param = ActionController::Parameters.new
=> <ActionController::Parameters {} permitted: false>
irb(main):009:0> param.permit(:a).to_h
=> {}
irb(main):010:0> param.permit(:a).to_hash
=> {}
```

connorshea: I probably made it nilable out of caution, you can remove it.

PR #113, opened 2019-08-13
Change activesupport safe_constantize to untyped

mgriisser:
`safe_constantize` returns either the value of the constant whose name is the same as the string it's called on or `nil` if that constant isn't defined. Since a constant can potentially have any value, `safe_constantize` should be untyped.


DarkDimius:
Thank you

PR #116, opened 2019-08-23
sorbet-runtime.rbi file to fix 4 lines in my "sorbet/rbi/todo.rbi" file

jasnow:
I have these 4 lines below in my "sorbet/rbi/todo.rbi" files so I created a sorbet-runtime.rbi file.
```
module T::InterfaceWrapper::Helpers; end
module T::Private::Abstract::Hooks; end
module T::Private::Methods::MethodHooks; end
module T::Private::Methods::SingletonMethodHooks; end
```
Part of https://github.com/sorbet/sorbet/issues/1488


jez:
fwiw i think this is probably not the right approach. todo.rbi *should* be just the constants that couldn’t resolve from type checking the code from a project. the fact that these are showing up in the todo.rbi means that either

- your project is referring to private, internal APIs
- the RBI generation scripts are generating references to private, internal APIs

Adding a sorbet-runtime.rbi to sorbet-typed is only going to paper over the problem, and make problems down the road. especially because there already is an RBI for sorbet-runtime here in Sorbet itself:

https://github.com/sorbet/sorbet/blob/master/rbi/sorbet/t.rbi

The true solution here would be to identify where the references to these constants are coming from (likely somewhere in the `sorbet/` folder) and change the RBI generation scripts to stop generating references to them.

jasnow:
>> "change the RBI generation scripts to stop generating references to them."

Does this mean that we must change the "sorbet" repo code?

jasnow:
Here is example of "git grep MethodHooks |sort -u": 
```
sorbet/rbi/gems/parlour.rbi:  extend T::Private::Methods::MethodHooks
sorbet/rbi/gems/parlour.rbi:  extend T::Private::Methods::SingletonMethodHooks
sorbet/rbi/gems/sorbet-rails.rbi:  extend T::Private::Methods::MethodHooks
sorbet/rbi/gems/sorbet-rails.rbi:  extend T::Private::Methods::SingletonMethodHooks
sorbet/rbi/todo.rbi:module T::Private::Methods::MethodHooks; end
sorbet/rbi/todo.rbi:module T::Private::Methods::SingletonMethodHooks; end
```

AaronC81:
This issue is occurring on my Parlour gem; I may be defining my abstract classes incorrectly, or there may be an issue with how Sorbet is generating types for them.

The generated `hidden.rbi` for an `abstract!` class seems to contain these two extends:
```ruby
extend ::T::Private::Abstract::Hooks
extend ::T::InterfaceWrapper::Helpers
```

They both appear in the `hidden.rbi` for the only two abstract classes in my project: https://github.com/AaronC81/parlour/blob/master/sorbet/rbi/hidden-definitions/hidden.rbi#L6225-L6233

I assume that their presence in `hidden.rbi` is then causing them to appear in `todo.rbi`.

jez:
@AaronC81 that makes sense, thanks for the context.

This is an implementation detail of how `abstract!` works. We should update `srb` to skip adding those `extend` lines to hidden.rbi.

jasnow:
>> "We should update srb to skip adding those extend lines to hidden.rbi."

Should I put this "should" into a separate GitHub issue?

jez:
To be clear I think we should not land this PR in it's current form, maybe at all. The solution instead would be to update `srb` to skip adding those extend lines to hidden.rbi.

jasnow:
@jez - I understand that this PR will not be merged. My question is how to track the "skip adding those extend lines to hidden.rbi" work.

PR #117, opened 2019-08-25
Separate migration methods by Rails version.

connorshea:
Create separate definitions for ActiveRecord::Migration::Compatibility::V5_2 and ActiveRecord::Migration::Compatibility::V6_0 so older migrations can still be typed.

Rails migrations look like this:

```ruby
# typed: true
class AddUserForeignKeyToExternalAccounts < ActiveRecord::Migration[5.2]
  def change
    add_foreign_key :external_accounts, :users, column: :user_id, on_delete: :cascade
  end
end
```

Rails migrations are supposed to be immutable, so Rails has compatibility classes for running older migrations in new versions of Rails. With 6.0, the existing ActiveRecord migration methods on Rails 5.2 migrations were no longer typed properly. This fixes that problem.

----

Because this deletes files - and if those files weren't deleted, things would probably break - I think we should wait until https://github.com/sorbet/sorbet/pull/1618 has been merged to merge this.

This changes a few things:

- Adds tests for ActiveRecord 5.2 and 6.0 specifically.
- Removes migration methods from ActiveRecord 'all', and moves the definitions into 5.2 and 6.0.
- Improves the migration class definitions for 5.2 and 6.0.
- Creates a `create_table` definition specifically for ActiveRecord 5.2 migrations on 6.0, since that's the only method that changed between the two versions (as far as I can tell).

The class names are different for each version, so I'm not sure how we can realistically share code between the separate version files. Unfortunately I think we just have to live with duplicate definitions for each version :/


connorshea:
@jez we can merge this now :)

jasnow:
I have version-specific migrations for 4.2, 5.0, 5.1, and 5.2. Will we support the others in the future?

PR #119, opened 2019-08-30
Add Gemfile file to repo

jasnow:
Add Gemfile file to repo to be more ruby-like.
Put Gemfile.lock in .gitignore because I assume it is only local.


Gemfile
```diff
@@ -0,0 +1,6 @@
+source 'https://rubygems.org'
+
+ruby '2.5.6'
+
+gem 'sorbet', :group => :development
+gem 'sorbet-runtime'
```
jez: We don't need to depend on sorbet-runtime.

Gemfile
```diff
@@ -0,0 +1,6 @@
+source 'https://rubygems.org'
+
+ruby '2.5.6'
+
+gem 'sorbet', :group => :development
```
jez: Can you edit the `.travis.yml` file to use this instead of manually installing gems?

https://github.com/sorbet/sorbet-typed/blob/master/.travis.yml

Gemfile
jasnow: Removed sorbet-runtime gem in below commit.

Gemfile
jasnow: Removed 2 lines from .travis.yml file.

PR #136, opened 2019-09-24
Add ActionView errors

panthomakos:
This change add some basic definitions for action view errors with
associated tests.

https://github.com/rails/rails/blob/master/actionview/lib/action_view/template/error.rb

### DIFF
--- a/lib/actionview/all/actionview.rbi
+++ b/lib/actionview/all/actionview.rbi
@@ -1,5 +1,23 @@
 # typed: strong
 
+module ActionView
+  class ActionViewError < StandardError; end
+  class EncodingError < StandardError; end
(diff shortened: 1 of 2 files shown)


lib/actionview/all/actionview.rbi
```diff
@@ -1,5 +1,23 @@
...
+  class MissingTemplate < ActionViewError
+    sig { returns(String) }
+    def path; end
+  end
+
+  class Template
+    class Error < ActionViewError; end
+  end
+
+  TemplateError = T.type_alias(Template::Error)
```
DarkDimius: This can be just `TemplateError = Template::Error`, they are ~equivalent for class aliases

PR #138, opened 2019-09-25
Add sigs for some `ActiveSupport::TimeWithZone` methods

ghiculescu:
ActiveSupport is massive, this barely scratches the surface. But it's a method I'd find useful, and have tested in my project.


lib/activesupport/all/activesupport_times_test.rb
```diff
@@ -0,0 +1,10 @@
+# typed: true
+
+require 'activesupport'
```
ghiculescu: not sure if this is needed?

ghiculescu:
sorry @DarkDimius there's one more issue here https://github.com/sorbet/sorbet-typed/pull/139

PR #140, opened 2019-09-25
Add interface for Rails view contexts

panthomakos:
Controllers in Rails have this method called `view_context` that returns an instance of a view context class. The basic view context is `ActionViews::Base`, but this can be configured with the `view_context_class`. The requirements for a view context class are that it can be initialized with three arguments, and that it responds to the `render` method with a `String`.

This change adds an interface for view contexts. A lot of the argument types are currently untyped. I believe they can be filled in later. The main thing this change provides is the proper interface for a view context, and the actual method that returns a view context so that Rails applications can type-check against it.

One question I have: is this the correct way to define an interface for this use case?

Another suggestion that is separate from this PR: when a new gem is added to this repository, it causes a lot of the hidden definitions or auto-generated files in a consuming repository to get blown away. While this makes sense, it does mean that the consuming repository is often left without placeholder method definitions and a lot of method missing type errors ensue. Would it be possible to provide a more additive solution? For example, maybe each directory could contain a `generated.rbi` file that we could remove methods from as we add more robust hand-written type definitions?

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -1,6 +1,7 @@
 # typed: strong
 
 class ActionController::Base < ActionController::Metal
+  include ActionView::Layouts
 end
 
 class AbstractController::Base < Object
--- a/lib/actionview/all/actionview.rbi
+++ b/lib/actionview/all/actionview.rbi
@@ -286,6 +286,7 @@ module ActionView::Helpers::UrlHelper
 end
 
 module ActionView::Layouts
+  include ActionView::Rendering
   extend T::Helpers  
 
   module ClassMethods ; end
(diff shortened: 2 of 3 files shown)


jez:
> A lot of the argument types are currently untyped. I believe they can be filled in later.

Honestly fine by me. I've pretty much never rejected a sorbet-typed PR because it had T.untyped.

> One question I have: is this the correct way to define an interface for this use case?

If this class doesn't exist at runtime, I'd suggest not putting it there. If you define a class in an RBI like this, people will try to do things like

```ruby
if x.is_a?(ViewContext)
  # ...
end
```

Sorbet won't complain about this statically, but it will blow up with a NameError at runtime.

Judging from my very limited assessment of the problem, I'd probably suggest tapping into something like sorbet-rails to generate something more accurate w.r.t. what the runtime is doing.

I'm not super familiar with Rails. I'd suggest joining our Slack (https://sorbet.org/slack) and joining the #rails channel, where there will be a bunch of people more experience with Rails to be able to answer your question.

If you discuss something there and decide that what you've implemented in this PR is the best way to proceed, that's fine by me. But probably best to have the discussion there first.

> Another suggestion that is separate from this PR:

I'm sorry, I didn't really follow this. At Stripe some of our tooling differs a little bit from the sorbet-typed-driven tooling, so we don't really feel the pains of using sorbet-typed as acutely as it seems you have. PRs to improve things are always welcome.

ptarjan:
Closing as we haven't heard back from the author. Feel free to reopen if we get an update.

PR #142, opened 2019-09-30
Added "include T::Enumerable" to fix sorbet/#1871

jasnow:
Added "include T::Enumerable" to fix sorbet/#1871.
Also Slack/Sorbet/Discuss: 
https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1569684429135200


aisamanra:
This won't quite suffice to fix the problem; I've gone ahead and done the subtle actual fix in #143, so I'm going to close this PR.

PR #145, opened 2019-10-02
Add very basic HTTParty types

ghiculescu:
This adds the bare minimum required to use `httparty` with `typed: true`. The rbi is for the latest `httparty` version, and the only change from what's auto-generated is online 416.

The test is just the example from https://github.com/jnunemaker/httparty#examples


ghiculescu:
Hmmm, do I need to do the same as https://github.com/sorbet/sorbet-typed/pull/141 ? Tests seem to pass either way.

aisamanra:
I think you will: it looks like that's a standing bug where we fail to catch it statically but it'll fail at runtime: https://github.com/sorbet/sorbet/issues/1805 (I think our tests here are only static.)

It should be easy enough to fix; do you want to add an `extend T::Helpers` to the relevant place?

ghiculescu:
Thanks @aisamanra that's done.

PR #146, opened 2019-10-02
Add T::Array[T::Boolean] to InWithinType

manhhung741:
We have a use-case in our code base that use [true, false] as value of the inclusion


manhhung741:
cc @connorshea

connorshea:
Should probably add a test, but this makes sense to me

manhhung741:
FYI I added a test for the change

manhhung741:
Can we merge this change?

PR #149, opened 2019-10-10
Fixed rake errors and upgrade ruby version in .travis.yml and Gemfile files

jasnow:
* Fixed rake errors in 2 rbi files.
 * Upgraded ruby version in .travis.yml and Gemfile files.


jasnow:
PS. The error messages were VERY CLEAR. I just followed them to fix the above files. THANKS for clear error messages.

DarkDimius:
> PS. The error messages were VERY CLEAR. I just followed them to fix the above files. THANKS for clear error messages.

Thank you, if you see some that are unclear, please let us know. We're striving to be super user friendly and that's one of the best ways we have.

PR #150, opened 2019-10-13
Add Rails::root

frewsxcv:
https://api.rubyonrails.org/classes/Rails.html#method-c-root


ghiculescu:
How do you feel about making this non-`nilable` @frewsxcv ?

I know it's theoretically possible for it to be nil, but that sounds like a pretty unlikely circumstance in everyday Rails use.

ghiculescu:
FYI: https://github.com/sorbet/sorbet-typed/pull/171

PR #152, opened 2019-10-15
Add advance to date and time

timgaleckas:
as extended by activesupport

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -539,6 +539,13 @@ class ActiveSupport::TimeWithZone
   def at_middle_of_day; end
 end
 
+# defines some of the methods at https://github.com/rails/rails/blob/v6.0.0/activesupport/lib/active_support/core_ext/date
+# this is not a complete definition!
+class Date
+  sig { params(options: T::Hash[Symbol, Integer]).returns(Date) }
+  def advance(options); end
+end
+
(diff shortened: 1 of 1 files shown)


lib/activesupport/all/activesupport.rbi
```diff
@@ -548,6 +556,10 @@ class Time
   sig { returns(Time) }
   def beginning_of_day; end
 
+  sig { params(options: T::Hash[Symbol, Integer]).returns(Time) }
+  def advance(options)
```
connorshea: These should be formatted as one line, like `def advance(options); end`

lib/activesupport/all/activesupport.rbi
```diff
@@ -539,6 +539,14 @@ class ActiveSupport::TimeWithZone
   def at_middle_of_day; end
 end
 
+# defines some of the methods at https://github.com/rails/rails/blob/v6.0.0/activesupport/lib/active_support/core_ext/date
+# this is not a complete definition!
+class Date
+  sig { params(options: T::Hash[Symbol, Integer]).returns(Date) }
+  def advance(options)
```
connorshea: This too

lib/activesupport/all/activesupport.rbi
timgaleckas: Yup. Fixed. Thanks.

lib/activesupport/all/activesupport.rbi
timgaleckas: Yup. Fixed. Thanks.

PR #153, opened 2019-10-16
Add ActiveRecord::Validations#{save,save!}

frewsxcv:
https://devdocs.io/rails~5.2/activerecord/validations#method-i-save

https://devdocs.io/rails~5.2/activerecord/validations#method-i-save-21


ghiculescu:
FYI, tweaking this in https://github.com/sorbet/sorbet-typed/pull/187

PR #155, opened 2019-10-20
fix signature for accepts_nested_attributes_for

manhhung741:
According to the API documentation and source code, this method accept a splash of attr_names
 
https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html#method-i-accepts_nested_attributes_for

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -203,7 +203,7 @@ end
 module ActiveRecord::NestedAttributes::ClassMethods
   sig do
     params(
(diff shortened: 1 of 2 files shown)


jez:
@manhhung741 looks like there are some test failures. Do they look related to your change?

manhhung741:
It's related to my change. I think it's because `sorbet` doesn't know `blank?` return T::Boolean here. I can fix it. I copied the test from Rails documentation^^

ptarjan:
Closing as we haven't heard back from the author. Feel free to reopen if we get an update.

PR #160, opened 2019-11-01
ActionController::Parameters#keys should take no arguments and return an array

pje:
```ruby
> ActionController::Parameters.new(foo: 'bar').keys
=> ["foo"]

> ActionController::Parameters.new(foo: 'bar').keys(:foo)
ArgumentError: wrong number of arguments (given 1, expected 0)
```

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -97,8 +97,8 @@ class ActionController::Parameters
   sig { params(key: T.any(String, Symbol)).returns(T::Boolean) }
   def key?(key); end
 
-  sig { params(key: T.any(String, Symbol)).returns(T.untyped) }
-  def keys(key); end
+  sig { returns(T.untyped) }
+  def keys; end
 
   sig { params(other_hash: T.untyped).returns(ActionController::Parameters) }
   def merge!(other_hash); end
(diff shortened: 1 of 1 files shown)


lib/actionpack/all/actionpack.rbi
```diff
@@ -97,8 +97,8 @@ class ActionController::Parameters
   sig { params(key: T.any(String, Symbol)).returns(T::Boolean) }
   def key?(key); end
 
-  sig { params(key: T.any(String, Symbol)).returns(T.untyped) }
-  def keys(key); end
+  sig { returns(T.untyped) }
+  def keys; end
```
DarkDimius: weren't arguments for this signature autogenerated?
Said otherwise, how sure are you that the sig is correct?

pje: > weren't arguments for this signature autogenerated?

I believe they were added by hand here: https://github.com/sorbet/sorbet-typed/pull/111

> are you sure that the changed argument list is correct?

If you provide more than 0 args to `keys`, it's an ArgumentError:

```ruby
> ActionController::Parameters.new(foo: 'bar').keys(:foo)
ArgumentError: wrong number of arguments (given 1, expected 0)
```

DarkDimius: Thank you!

Issue #159, opened 2019-10-30
VariadicUntypedFunction for a scope with (date = DateTime.now)


mikkeljuhl:
I have a rails model which has a scope: 

`scope :ready, -> (date = DateTime.now) { where('date < ?', date) }`

I get the following error: 

```
Expected T.any(T.proc.returns(T.untyped), T.proc.params(arg0: T.proc.returns(T.untyped)).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped, arg7: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped, arg7: T.untyped, arg8: T.untyped).returns(T.untyped), T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped, arg7: T.untyped, arg8: T.untyped, arg9: T.untyped).returns(T.untyped)) but found Proc for argument body https://srb.help/7002
    44 |  scope :ready, -> (date = DateTime.now) { where('date < ?', date) }
[...]

PR #161, opened 2019-11-04
Turn VariadicUntypedFunction type alias into `Proc`

mikkeljuhl:
See #159 for justification

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -1,21 +1,6 @@
 # typed: strong
 
-VariadicUntypedFunction = T.type_alias do
-  T.any(
-    T.proc.returns(T.untyped),
-    T.proc.params(arg0: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped).returns(T.untyped),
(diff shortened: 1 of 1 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -1,21 +1,6 @@
...
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped, arg7: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped, arg7: T.untyped, arg8: T.untyped).returns(T.untyped),
-    T.proc.params(arg0: T.untyped, arg1: T.untyped, arg2: T.untyped, arg3: T.untyped, arg4: T.untyped, arg5: T.untyped, arg6: T.untyped, arg7: T.untyped, arg8: T.untyped, arg9: T.untyped).returns(T.untyped)
-    # Currently Sorbet is limited to procs with 9 args: https://github.com/sorbet/sorbet/blob/df6085e6ae9846f033064513ea14f069c68b0bf9/core/SymbolRef.h#L387
-  )
-end
+VariadicUntypedFunction = T.type_alias { Proc }
```
jez: @mikkeljuhl i think an even better long term fix here is do replace all **usages** of VariadicUntypedFunction with Proc. This type alias isn't defined at runtime, so it will fail at runtime and not statically if people try to use this type alias in their real code (not RBIs)

PR #163, opened 2019-11-11
Add ActiveSupport Numeric time extensions.

frewsxcv:
Everything in this file:

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/numeric/time.rb

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -616,3 +616,44 @@ class Hash
   sig { returns(T::Hash[Symbol, T.untyped]) }
   def to_options; end
 end
+
+class Numeric
+  sig { returns(ActiveSupport::Duration) }
+  def second; end
+
+  sig { returns(ActiveSupport::Duration) }
+  def seconds; end
+
+  sig { returns(ActiveSupport::Duration) }
+  def minute; end
+
+  sig { returns(ActiveSupport::Duration) }
+  def minutes; end
+
+  sig { returns(ActiveSupport::Duration) }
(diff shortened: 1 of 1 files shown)


ghiculescu:
This doesn't play nicely with `sorbet-rails` for me. For example, if you have an integer column, `sorbet-rails` generates this:

```ruby
  sig { params(value: T.nilable(Integer)).void }
  def update_interval=(value); end
```

With this update, this code no longer works:

```ruby
  foo.update_interval = 1.day
```

Instead it gets `Assigning a value to value that does not match expected type T.nilable(Integer)`.

Assuming you use Rails, how are you handling this @frewsxcv ?

ghiculescu:
My best idea so far is to change https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/model_plugins/active_record_attribute.rb#L95 to return multiple types, so it would return `T.any(Integer, ActiveSupport::Duration)`.

Or to add a `.to_i` in the code that's erroring.

elliottt:
Should we revert this?

frewsxcv:
> My best idea so far is to change https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/model_plugins/active_record_attribute.rb#L95 to return multiple types, so it would return `T.any(Integer, ActiveSupport::Duration)`.

Agreed. This should be fixed in sorbet-rails 👍

ghiculescu:
Cool, working on an improvement here: https://github.com/chanzuckerberg/sorbet-rails/pull/223

PR #166, opened 2019-11-12
Add extremely minimal `graphql` types

ghiculescu:
This is (only just) enough to get you started with https://github.com/rmosolgo/graphql-ruby

The main reason for this is because `srb rbi gems` doesn't include the `GraphQL::Schema::Member::HasArguments` include, which means you can't use arguments out of the box. There's probably other really entry level features that could easily be added here too.

### DIFF
--- a/lib/graphql/all/graphql.rbi
+++ b/lib/graphql/all/graphql.rbi
@@ -1,0 +1,16 @@
+# typed: true
+
+module GraphQL::Schema::Member::HasFields
+  def field(*args, **kwargs, &block); end
+end
+
+module GraphQL::Schema::Member::HasArguments
+  def argument(*args, **kwargs, &block); end
+end
+
+class GraphQL::Schema::Object < GraphQL::Schema::Member
+  extend GraphQL::Schema::Member::HasFields
+
+  # srb rbi gems does *not* add the next line, which means arguments don't work out of the box
+  extend GraphQL::Schema::Member::HasArguments
+end
--- a/lib/graphql/all/graphql_test.rb
+++ b/lib/graphql/all/graphql_test.rb
@@ -1,0 +1,12 @@
+# typed: true
+
+module GraphqlTest
+  class User < GraphQL::Schema::Object
+  end
+
+  class Query < GraphQL::Schema::Object
+    field :user, User, "The current user", null: true do
+      argument :id, ID, "The ID of the user to find", required: true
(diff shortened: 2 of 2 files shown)


connorshea:
I came to essentially the same conclusion when trying to typecheck graphql-ruby for my codebase a few weeks ago. Though, my solution is much stupider :)

```ruby
# Dumb hack to work around an issue in Sorbet with nilable blocks not
# letting you use `T.proc.bind`.
# https://github.com/sorbet/sorbet/issues/498
# This allows `argument` to be used within field blocks, even though the
# actual solution should look more like this:
#
# ```ruby
# module GraphQL::Schema::Member::HasFields
#   sig do
#     params(
#       args: T.untyped,
#       kwargs: T.untyped,
#       block: T.nilable(T.proc.bind(GraphQL::Schema::Field).void)
#     ).returns(T.untyped)
#   end
#   def field(*args, **kwargs, &block); end
# end
# ```
#
# Note that this will only impact QueryType, and will need to be done for every
# class where we want to use fields with arguments.
class Types::QueryType < Types::BaseObject
  extend GraphQL::Schema::Member::HasArguments
end
```

I should note that the `extend GraphQL::Schema::Member::HasArguments` is incorrect, that's never actually extended in the graphql-ruby codebase, it should be bound via the block parameter on `field`, but as mentioned in my code comments that's not currently possible due to https://github.com/sorbet/sorbet/issues/498.

ghiculescu:
> I should note that the `extend GraphQL::Schema::Member::HasArguments` is incorrect, that's never actually extended in the graphql-ruby codebase, it should be bound via the block parameter on `field`, but as mentioned in my code comments that's not currently possible due to [sorbet/sorbet#498](https://github.com/sorbet/sorbet/issues/498).

Yes, that's true. I'm happy to change it if you can think of something that's more correct, but also works :) 

In the interim how do you feel about adding a comment noting this to https://github.com/sorbet/sorbet-typed/pull/168/files so that future readers of my crappy code are aware of this conversation? @connorshea

manhhung741:
> It should be bound via the block parameter on field, but as mentioned in my code comments that's not currently possible due to sorbet/sorbet#498.

Yes, this will continue to be a big issue for writing proper signatures for libraries like Rails & GraphQL. Even generating signatures don't work if we don't have support for features like this in Sorbet.

PR #168, opened 2019-11-13
Add value method for GraphQL enums.

connorshea:
Partially generated by Sord, with some modifications.

This is building off of #166.

https://graphql-ruby.org/type_definitions/enums.html


lib/graphql/all/graphql.rbi
```diff
@@ -14,3 +14,19 @@ class GraphQL::Schema::Object < GraphQL::Schema::Member
   # srb rbi gems does *not* add the next line, which means arguments don't work out of the box
```
ghiculescu:
```suggestion
  # srb rbi gems does *not* add the next line, which means arguments don't work out of the box
  # Note this isn't strictly correct. see https://github.com/sorbet/sorbet-typed/pull/166
```

PR #164, opened 2019-11-11
Add ActiveRecord::Relation#delete_all

frewsxcv:
https://devdocs.io/rails~5.2/activerecord/relation#method-i-delete_all

PR #169, opened 2019-12-02
`ActiveRecord::Relation` is a class, not a module

ghiculescu:
fixes a bug in https://github.com/sorbet/sorbet-typed/pull/164


frewsxcv:
Woooops, thanks for fixing this!

PR #176, opened 2019-12-17
Add definitions for `end_of_day` and its alias `at_end_of_day`

DanielGilchrist:
Adds definitions for `end_of_day` and its alias `at_end_of_day` for `ActiveSupport::TimeWithZone` from https://github.com/rails/rails/blob/v6.0.0/activesupport/lib/active_support/core_ext/time/calculations.rb#L219


jez:
Thanks!

PR #179, opened 2020-01-05
Add a skeleton type signature for HashWithIndifferentAccess

connorshea:
This adds a basic type signature for [`HashWithIndifferentAccess`](https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html), which is a Rails class that descends from `Hash`, but with the ability to access hash members via either String or Symbol keys.

See also `hash_with_indifferent_access.rb`: https://github.com/rails/rails/blob/f33d52c95217212cbacc8d5e44b5a8e3cdc6f5b3/activesupport/lib/active_support/hash_with_indifferent_access.rb

It only uses enough of the generics system to pass typechecking, so it doesn't really have fully correct type signatures yet. I still need to add tests as well.

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -812,3 +812,338 @@ class ActiveSupport::Duration
   sig { params(precision: T.nilable(Integer)).returns(String) }
   def iso8601(precision: nil); end
 end
+
+# Implements a hash where keys `:foo` and `"foo"` are considered
+# to be the same.
+#
+# ```ruby
+# rgb = ActiveSupport::HashWithIndifferentAccess.new
+#
+# rgb[:black] = '#000000'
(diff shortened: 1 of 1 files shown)


lib/activesupport/all/activesupport.rbi
```diff
@@ -812,3 +812,338 @@ class ActiveSupport::Duration
...
+# which will, in turn, require this file.
+class ActiveSupport::HashWithIndifferentAccess < Hash
+  include Enumerable
+
+  extend T::Generic
+  K = type_member
+  V = type_member
+  Elem = type_member
+
+  sig { params(key: T.any(String, Symbol)).returns(T.untyped) }
```
ghiculescu: suggestion: make an internal type alias for `T.any(String, Symbol)` so you don't accidentally only allow one or the other on any method

PR #181, opened 2020-01-13
AR association callbacks accept an array

ghiculescu:
See docs @ https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html

image

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -7,8 +7,9 @@ AssociationCallback = T.type_alias do
     T.any(
       Symbol, # reference to a method
       String, # reference to a method? not clear: no string examples in docs
-      T.proc.void, # a proc that contains the callback
-      T::Array[T.any(Symbol, Proc)] # multiple callbacks
+      T.proc.void, # a lambda that contains the callback
+      Proc, # a proc that contains the callback
+      T::Array[T.any(Symbol, Proc, T.proc.void)] # multiple callbacks
     )
   )
 end
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -1,19 +1,30 @@
 # typed: strong
 
 VariadicUntypedFunction = T.type_alias { Proc }
+AssociationCallback = T.type_alias do
+  # docs in https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html - jump to "Association callbacks"
+  T.nilable(
+    T.any(
+      Symbol, # reference to a method
+      String, # reference to a method? not clear: no string examples in docs
+      T.proc.void, # a proc that contains the callback
+      T::Array[T.any(Symbol, Proc)] # multiple callbacks
+    )
+  )
+end
 
 module ActiveRecord::Associations::ClassMethods
   sig do
     params(
       name: Symbol,
       scope: T.nilable(T.proc.void),
-      after_add: T.nilable(T.any(Symbol, String, T.proc.void)),
-      after_remove: T.nilable(T.any(Symbol, String, T.proc.void)),
+      after_add: AssociationCallback,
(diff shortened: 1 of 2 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -1,19 +1,30 @@
...
 
 VariadicUntypedFunction = T.type_alias { Proc }
+AssociationCallback = T.type_alias do
+  # docs in https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html - jump to "Association callbacks"
+  T.nilable(
+    T.any(
+      Symbol, # reference to a method
+      String, # reference to a method? not clear: no string examples in docs
+      T.proc.void, # a proc that contains the callback
+      T::Array[T.any(Symbol, Proc)] # multiple callbacks
```
ghiculescu: i dont really understand why `Proc` works here but `T.proc.void` doesn't, [same issue here](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Breturns(T%3A%3AArray%5BT.any(Symbol%2C%20T.proc.void)%5D)%7D%0Adef%20bar%0A%20%20%5B%3Afoo%2C%20Proc.new%20%7B%20%7D%5D%0Aend%0A%0Asig%20%7Breturns(T%3A%3AArray%5BT.any(Symbol%2C%20Proc)%5D)%7D%0Adef%20bar2%0A%20%20%5B%3Afoo%2C%20Proc.new%20%7B%20%7D%5D%0Aend%0A)

elliottt: Do you get a type error in the rbi file, or does it end up being impossible to use?

ghiculescu:
```
testing activerecord all
activerecord_test.rb:22: Expected T.nilable(T.any(Symbol, String, T.proc.void, T::Array[T.any(T.proc.void, Symbol)])) but found [Symbol(:"foo"), Proc] for argument after_add https://srb.help/7002
    22 |  has_and_belongs_to_many :projects, after_add: [:foo, Proc.new { }]
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    activerecord.rbi:158: Method ActiveRecord::Associations::ClassMethods#has_and_belongs_to_many has specified after_add as T.nilable(T.any(Symbol, String, T.proc.void, T::Array[T.any(T.proc.void, Symbol)]))
     158 |      after_add: AssociationCallback,
                ^^^^^^^^^
  Got [Symbol(:"foo"), Proc] originating from:
    activerecord_test.rb:22:
    22 |  has_and_belongs_to_many :projects, after_add: [:foo, Proc.new { }]
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

elliottt: I think the problem is with the use of `Proc.new {}`. If you change that to `lambda {}` or `-> {}` it works fine.

ghiculescu: Ahh gotcha. I just copied the example from the Rails docs. Maybe I should just do `T::Array[T.any(Symbol, Proc, T.proc.void)]`?

ghiculescu: Should we add some more detail to https://sorbet.org/docs/procs about the difference between `Proc` and `T.proc`?

elliottt: Sure! Do you feel like taking an initial pass at that?

ghiculescu: It's on the todo list :) In the meantime I've fixed this one up.

elliottt: Awesome, thank you!

PR #182, opened 2020-01-14
first touch rest-client

AlexMouton:
Hi just getting started here with a rest-client stub. 
Hope this helps.
Alex

### DIFF
--- a/lib/rest-client/all/rest-client.rbi
+++ b/lib/rest-client/all/rest-client.rbi
@@ -1,0 +1,3 @@
+class RestClient::Request
+    def self.execute(args, &blk); end
+end


lib/rest-client/all/rest-client.rbi
```diff
@@ -0,0 +1,3 @@
+class RestClient::Request
+    def self.execute(args, &blk); end
```
elliottt: Could you add a type signature here?

AlexMouton: I am just starting to get sorbet into a big codebase. I can get back to that once I have decent types elsewhere.

AlexMouton:
Im going to close this mr until I know better which external code goes here and which can be generated from the gem itself.

PR #183, opened 2020-01-14
First touch pg

AlexMouton:
Hi just getting started here with a pg stub. 
Hope this helps.
Alex

### DIFF
--- a/lib/pg/all/pg.rbi
+++ b/lib/pg/all/pg.rbi
@@ -1,0 +1,3 @@
+class PG 
+    def self.connect(params); end
+end


ghiculescu:
This is causing errors:

```
sorbet/rbi/sorbet-typed/lib/pg/all/pg.rbi:8: PG was previously defined as a module https://srb.help/4012
     8 |class PG
     9 |    def self.connect(params); end
    10 |end
```

Probably because the file generated by `srb rbi gems` starts with this:

```
module PG
  def self.connect(*args); end
  def self.init_openssl(arg0, arg1); end
  def self.init_ssl(arg0); end
  def self.is_threadsafe?; end
  def self.isthreadsafe; end
  def self.library_version; end
  def self.threadsafe?; end
  def self.version_string(include_buildnum = nil); end
end
```

Can we revert?

jez:
I've reverted this PR in #185. Thanks for flagging @ghiculescu

PR #184, opened 2020-01-15
Normalized string singularize to Inflector.singularize.

AlexMouton:
String.singularize is implemented in terms of Inflector.singularize and so their signatures should be equal.

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -189,7 +189,7 @@ class String
   sig { returns(T.untyped) }
   def safe_constantize; end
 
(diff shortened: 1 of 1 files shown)


elliottt:
Is there an example of when this method will return `nil`?

ghiculescu:
@elliottt the nilable usage looks like a bug in the sig. This never returns nil.

elliottt:
Oh! I misread the diff. I thought this was making the return type nilable. I've been reading too many PRs today, thank you for pointing that out!

ghiculescu:
All good. Thanks for powering through all these <3

PR #190, opened 2020-01-30
Date#- should work with AS::Duration

ghiculescu:
There's been a bunch of improvements to the sig for `Date#-` recently: https://github.com/sorbet/sorbet/pull/2514 https://github.com/sorbet/sorbet/pull/2539 https://github.com/sorbet/sorbet/pull/2567 https://github.com/sorbet/sorbet/pull/2570

This PR adds one more sig, for subtracting an `ActiveSupport::Duration`. Annoyingly ActiveSupport::Duration is not a subclass of `Numeric` even though it more or less behaves like one. So the sigs included in sorbet didn't work anymore.

@elliottt @aisamanra FYI, tagging you since you both made improvements to the `Date#-` sigs (thank you!).

### DIFF
--- a/lib/activesupport/all/activesupport_duration_test.rb
+++ b/lib/activesupport/all/activesupport_duration_test.rb
@@ -63,5 +63,8 @@ module ActiveSupportDurationTest
   T.assert_type!(3.years % 5, ActiveSupport::Duration)
   T.assert_type!(3.years % 5.days, ActiveSupport::Duration)
 
+  # see https://github.com/sorbet/sorbet-typed/pull/190
   T.assert_type!(Date.today - 4.weeks, Date)
+  T.assert_type!(Date.today - Date.today, Rational)
+  T.assert_type!(Date.today - 4, Date)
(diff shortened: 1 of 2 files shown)


lib/activesupport/all/activesupport_duration_test.rb
```diff
@@ -62,4 +62,9 @@ module ActiveSupportDurationTest
 
   T.assert_type!(3.years % 5, ActiveSupport::Duration)
   T.assert_type!(3.years % 5.days, ActiveSupport::Duration)
+
+  # see https://github.com/sorbet/sorbet-typed/pull/190
+  T.assert_type!(Date.today - 4.weeks, Date)
+  T.assert_type!(Date.today - Date.today, Rational)
```
ghiculescu: this test requires a relatively recent version of sorbet to pass (it should match https://github.com/sorbet/sorbet/pull/2567). i'm not sure how CI works with that - i guess we're about to find out.

PR #192, opened 2020-01-30
Time#- should work with AS::Duration

ghiculescu:
same as https://github.com/sorbet/sorbet-typed/pull/190 but for `Time`


ghiculescu:
oops hold off, tests will be red

ghiculescu:
okay should be good now.

PR #199, opened 2020-02-02
Refine `Time#in_time_zone` and `String#in_time_zone` sigs

DanielGilchrist:
`in_time_zone` on the `Time` and `String` classes allow `nil` to be passed as an argument for `zone`

```ruby
# returns a Time object
Time.now.in_time_zone(nil)
# => 2020-02-02 11:30:51 +0000
```
```ruby
# returns a Time object
Time.now.to_s.in_time_zone(nil)
# => 2020-02-02 11:32:10 +0000
```

Additionally updated the signatures to accept ~either~ an `ActiveSupport::TimeZone` ~or a `String` and updated their return types to be strictly `ActiveSupport::TimeWithZone`. This is because `Time` is only returned in the case `nil` is passed.~

I'm not too sure what the go is when it comes to multiple signatures in RBI files but let me know if there are any issues! Cheers :)

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -154,6 +154,7 @@ class String
   def humanize(capitalize: true, keep_id_suffix: false); end
 
   sig { params(zone: String).returns(T.any(Time, ActiveSupport::TimeWithZone)) }
+  sig { params(zone: NilClass).returns(Time) }
   def in_time_zone(zone = ::Time.zone); end
 
   sig { params(amount: Integer, indent_string: T.nilable(String), indent_empty_lines: T::Boolean).returns(T.nilable(String)) }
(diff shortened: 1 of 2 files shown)


DanielGilchrist:
> We don't allow overloaded method signatures in RBI files outside of Sorbet's internal payload itself.
> 
> (It's a bug that Sorbet doesn't currently report an error on your PR causing the tests to fail, but I'm dealing with that separately.)
> 
> Can you change this to avoid using overloads? If you find yourself stuck, in general we opt for less type safety with simpler sigs over more type safety at the cost of complex sigs (overloads being complex).

No worries, totally fair and thanks for the info! I can just update the signatures to accept `NilClass` instead 🙂 Would you prefer I close this PR and make a new one with a clean commit history?

jez:
> Would you prefer I close this PR and make a new one with a clean commit history?

We will squash your changes before merging; use whatever commit history you feel like.

ghiculescu:
> (It's a bug that Sorbet doesn't currently report an error on your PR causing the tests to fail, but I'm dealing with that separately.)

Does that mean that usage of multiple sigs in user code is going to start erroring at some point in the future? If so we (and probably many others) need to start preparing for that.

If/when that happens, does the team have any advice on how to deal with monkey patch scenarios like https://github.com/sorbet/sorbet-typed/pull/192 ?

PR #201, opened 2020-02-04
Add the `prepend` option to ActiveRecord's `before_destroy`

bradleybuda:
Documented here: https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html

This has been around since Rails 3, so adding to the "all" version

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -460,13 +460,15 @@ class ActiveRecord::Base
     params(
       arg: T.nilable(Symbol),
       if: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean)))),
-      unless: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean))))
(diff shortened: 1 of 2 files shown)


lib/activerecord/all/activerecord.rbi
```diff
@@ -460,13 +460,15 @@ class ActiveRecord::Base
     params(
       arg: T.nilable(Symbol),
       if: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean)))),
-      unless: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean))))
+      unless: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean)))),
+      prepend: T.nilable(T::Boolean)
```
jez: Will it actually accept `before_destroy :foo, prepend: nil`? Do you want to allow that?

i.e., it might be better to take a non-nilable `T::Boolean`, and set the default to `false` instead of `nil`.

jez:
One comment, feel free to address in a follow up PR if you care about it.

PR #214, opened 2020-02-19
add `add_belongs_to` support to activerecord migrations

deecewan:
simple one, but there are a lot of aliases that people use 🤷‍♂


lib/activerecord/~>5.2.0/activerecord.rbi
```diff
@@ -376,6 +376,8 @@ class ActiveRecord::Migration::Current < ActiveRecord::Migration
     polymorphic: false,
     null: nil
   ); end
+  
+  alias add_belongs_to add_reference
```
jez: Note that it's likely that in the future that methods defined with `alias` will all be as if the new method was given no sig. I'd encourage you to duplicate the RBI definition for `add_reference` instead of using `alias`.

PR #218, opened 2020-02-21
Add ActiveModel::Errors methods.

connorshea:
And errors method on ActiveModel::Validations, plus some tests.

[`ActiveModel::Errors`](https://api.rubyonrails.org/classes/ActiveModel/Errors.html) is used to handle errors when validating models in Rails. This adds some of the most common methods for this :)

### DIFF
--- a/lib/activemodel/all/activemodel_test.rb
+++ b/lib/activemodel/all/activemodel_test.rb
@@ -77,3 +77,24 @@ module ActiveModelTest
   validates_numericality_of :width, greater_than: :minimum_weight, on: [:create, :update]
   validates_presence_of :first_name, on: [:create, :update]
 end
+
+class Person
+  include ActiveModel::Validations
+end
+
+class ActiveModelErrorsTest
+  person = Person.new
+  person.errors.add(:name)
+  person.errors.add(:name, :invalid, strict: true)
+  person.errors.add(:name, :invalid, strict: NameIsInvalid)
+  person.errors.add(:name, :too_long, { count: 25 })
+  person.errors.add(:base, :name_or_email_blank, message: "either name or email must be present")
+  person.errors.of_kind?(:age)
+  person.errors.of_kind?(:name, :too_long)
+  person.errors.of_kind?(:name, "is too long (maximum is 25 characters)")
+  person.errors.full_messages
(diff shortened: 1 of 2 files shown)


lib/activemodel/all/activemodel_test.rb
```diff
@@ -77,3 +77,24 @@ module ActiveModelTest
   validates_numericality_of :width, greater_than: :minimum_weight, on: [:create, :update]
   validates_presence_of :first_name, on: [:create, :update]
 end
+
+class Person
```
ghiculescu: can you please add a test where this descends from `ActiveRecord::Base` too?

there were issues with the `ActiveModel::Validations` sigs where they didn't work properly in AR, which a test would have caught - see https://github.com/sorbet/sorbet-typed/pull/216/files#diff-52d2b7591f89a773def6fce9c1ca7ad7R80 for an example.

connorshea: You mean `ApplicationRecord`? I can, but it causes Person not to have a `new` method so I have to write the initializer like `person = T.let(T.unsafe(nil), Person)`

DarkDimius:
@ghiculescu should we merge this?

ghiculescu:
This is fine to merge, it just may not actually do what's expected without the extra test I suggested.

lib/activemodel/all/activemodel_test.rb
ghiculescu: @connorshea did you also add this line? https://github.com/sorbet/sorbet-typed/pull/216/files#diff-52d2b7591f89a773def6fce9c1ca7ad7R39

connorshea: @ghiculescu that link doesn't actually refer to a specific line.

ghiculescu: Indeed it doesn't, and now I can't remember what I was trying to link to :(

ghiculescu: Oh right. It was this line: https://github.com/sorbet/sorbet-typed/blob/master/lib/activerecord/all/activerecord_test.rb#L39

PR #220, opened 2020-02-21
Fix optional arg in Grape

ghiculescu:
https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/inside_route.rb#L145

- fixes an optional arg
- this method always throws, so `T.noreturn` is appropriate


connorshea:
Looks fine to me.

DarkDimius:
Thank you

PR #225, opened 2020-03-16
Make constantize return a class

wakeless:
A Module is a Class so Class is the most correct return type

It raises if it isn't in existence, so not nilable

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -119,7 +119,7 @@ class String
   sig { returns(String) }
   def classify; end
 
-  sig { returns(Class) }
+  sig { returns(Module) }
   def constantize; end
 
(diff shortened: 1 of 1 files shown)


lib/activesupport/all/activesupport.rbi
```diff
@@ -119,7 +119,7 @@ class String
   sig { returns(String) }
   def classify; end
 
-  sig { returns(String) }
+  sig { returns(Class) }
```
paracycle: This is not quite correct:
```
[1] pry(main)> Module < Class
=> false
[2] pry(main)> Class < Module
=> true
```
This return should be `Module`.

paracycle:
```suggestion
  sig { returns(Module) }
```

PR #226, opened 2020-03-18
Add signature for ActiveSupport::TimeWithZone#midnight?

DanielGilchrist:
```ruby
Time.zone.now.class
# => ActiveSupport::TimeWithZone < Object
Time.zone.now.midnight?
#=> false
```


DanielGilchrist:
@elliottt I just realised this is monkey patched in our code base OOPS could you please revert it? Sorry about that!!

PR #231, opened 2020-04-01
add `array` option to `add_column`

deecewan:
As shown [here](https://github.com/rails/rails/blob/v5.2.3/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L572)


elliottt:
Thanks!

PR #235, opened 2020-04-09
Add rbi for ActiveRecord::Sanitization::ClassMethods

dikond:


dikond:
Thanks for fast response!

PR #246, opened 2020-05-04
Add `ActiveSupport::TimeWithZone#advance`

ghiculescu:
eg. for `Time.zone.now.advance(hours: 2)`

https://api.rubyonrails.org/v6.0.2.1/classes/ActiveSupport/TimeWithZone.html#method-i-advance

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -894,7 +894,7 @@ class ActiveSupport::TimeWithZone
   # The options parameter takes a hash with any of these keys: :years, :months, :weeks, :days, :hours, :minutes, :seconds.
   # If advancing by a value of variable length (i.e., years, weeks, months, days), move forward from `time`, otherwise move forward
   # from utc, for accuracy when moving across DST boundaries.
-  sig { params(options: T::Hash[Symbol, Integer]).returns(ActiveSupport::TimeWithZone) }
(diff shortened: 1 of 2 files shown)


lib/activesupport/all/activesupport.rbi
```diff
@@ -888,6 +888,14 @@ class ActiveSupport::TimeWithZone
   # of `ActiveSupport.to_time_preserves_timezone`.
   sig { returns(Time) }
   def to_time; end
+
+  # Uses Date to provide precise Time calculations for years, months, and days according to the proleptic Gregorian calendar.
+  # The result is returned as a new `TimeWithZone` object.
+  # The options parameter takes a hash with any of these keys: :years, :months, :weeks, :days, :hours, :minutes, :seconds.
+  # If advancing by a value of variable length (i.e., years, weeks, months, days), move forward from `time`, otherwise move forward
+  # from utc, for accuracy when moving across DST boundaries.
+  sig { params(options: T::Hash[Symbol, Integer]).returns(ActiveSupport::TimeWithZone) }
```
elliottt, line 897: Does this only work with `Integer`, or do `Float` values work as well

ghiculescu: Looks like either works.

elliottt:
Thanks!

PR #248, opened 2020-05-18
Add shellwords.rbi

wakeless:
Part of the stdlib. Used the documentation as a guide.

https://ruby-doc.org/stdlib-2.5.1/libdoc/shellwords/rdoc/Shellwords.html#method-c-escape


jez:
Since this is part of the stdlib, would you mind sending this to https://github.com/sorbet/sorbet/tree/master/rbi/stdlib ?

wakeless:
Yep, thanks for quick review.

PR #250, opened 2020-05-26
Add ActiveRecord::Transactions::ClassMethods after_create_commit definition

caiofilipemr:
Adding [after_create_commit](https://apidock.com/rails/v5.0.0.1/ActiveRecord/Transactions/ClassMethods/after_create_commit) definition for Rails 5+

### DIFF
--- a/lib/activerecord/~>5/activerecord.rbi
+++ b/lib/activerecord/~>5/activerecord.rbi
@@ -1,4 +1,4 @@
-# typed: strong
+# typed: strict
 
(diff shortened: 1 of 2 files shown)


lib/activerecord/~>5/activerecord.rbi
```diff
@@ -0,0 +1,17 @@
+# typed: strong
```
jez, line 1:
```suggestion
# typed: strict
```

PR #257, opened 2020-06-04
Remove RBI for Bundler as it's now part of Sorbet's payload

Morriar:
As Bundler is [part of Ruby's standard library since 2.6](http://ruby-doc.com/stdlib-2.6.5/libdoc/bundler/rdoc/Bundler.html), I moved it to Sorbet's payload here: https://github.com/sorbet/sorbet/pull/3135

This PR removes the `bundler.rbi` as it will be included in future versions of Sorbet.


jez:
This change lgtm but i'm going to hold off until the upstream PR lands.

Morriar:
This one will need to be merged before as it's blocking https://github.com/sorbet/sorbet/pull/3135

PR #258, opened 2020-06-04
Remove open3.rbi as all the signatures are already in Sorbet's payload

Morriar:
All these signatures seemed redundant with the ones in Sorbet's payload for [open3.rbi](https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/open3.rbi).


jez:
thanks!

PR #259, opened 2020-06-04
Remove resolv.rbi signatures that belong to Sorbet's payload

Morriar:
As `resolv` is part of the standard library, I moved it to Sorbet's payload here: https://github.com/sorbet/sorbet/pull/3150

This PR removes the `resolv.rbi` as it will be included in future versions of Sorbet.


jez:
this lgtm, but i'm going to hold off merging until the upstream change lands.

Morriar:
Since sorbet/sorbet#3150 has been merged, I think we can safely merge this one.

PR #268, opened 2020-07-14
Add faraday.rbi

dduugg:
This builds on the autogenerated one, which didn't create sigs for the basic case (e.g. `Faraday.get`) because the gem [proxies](https://github.com/lostisland/faraday/blob/v1.0.1/lib/faraday.rb#L112-L121) HTTP verb methods. This is done in a separate commit for ease of review.

Tests are sourced from https://lostisland.github.io/faraday/usage/ (However, I'm noticing that tests pass even with code that should fail. I've posted details in Slack.)


elliottt:
👋  I think the problem with the tests is that it's named `faraday_test.rbi` instead of just `faraday_test.rb`. Renaming should get them running.

dduugg:
@elliottt 🤦 
I've renamed the file and cleaned things up a bit, ptal

PR #269, opened 2020-07-15
Create sucker_punch.rbi

okhwaja:
To get my usages of suckerpunch to type check correct, I needed to manually add `mixes_in_class_methods(SuckerPunch::Job::ClassMethods)` into the module that gets included.

Normally it's used like so

```ruby
class MyJob
  include SuckerPunch::Job
  
  def perform(job_data)
  end
end
```


lib/sucker_punch/all/sucker_punch.rbi
```diff
@@ -0,0 +1,89 @@
...
+class SuckerPunch::Counter::Failed
+  def counter; end
+  def initialize(queue_name); end
+  def self.clear; end
+  include SuckerPunch::Counter::Utilities
+end
+module SuckerPunch::Job
+  def logger; end
+  def self.included(base); end
+  mixes_in_class_methods(SuckerPunch::Job::ClassMethods)
```
okhwaja, line 49: this line is the only addition - rest was autogenerated

lib/sucker_punch/all/sucker_punch.rbi
```diff
@@ -0,0 +1,89 @@
+# This file is autogenerated. Do not edit it by hand. Regenerate it with:
+#   srb rbi gems
+
+# typed: true
+#
+# If you would like to make changes to this file, great! Please create the gem's shim here:
+#
+#   https://github.com/sorbet/sorbet-typed/new/master?filename=lib/sucker_punch/all/sucker_punch.rbi
+#
```
elliottt, line 9: Would you mind removing the autogenerated header?

PR #272, opened 2020-07-19
Fix some active support array types

DanielGilchrist:
These methods access with a hard coded index using `Array#[]` which results in `nil` if the element can't be found. Also fixed a typo where `to_default_s` was defined as `to_defaul_s`

Example in docs: https://api.rubyonrails.org/classes/Array.html#method-i-second
`to_default_s`: https://api.rubyonrails.org/classes/Array.html#method-i-to_default_s

### DIFF
--- a/lib/activesupport/all/activesupport.rbi
+++ b/lib/activesupport/all/activesupport.rbi
@@ -257,7 +257,7 @@ class Array
   sig { returns(T.nilable(Elem)) }
   def forty_two; end
 
-  sig { returns(Elem) }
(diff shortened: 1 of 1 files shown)


ghiculescu:
i think you missed `fourth` https://github.com/sorbet/sorbet-typed/pull/272/files#diff-cb18be98aefd3385e8588bf7600c9fe9R261

DanielGilchrist:
> Looks good, did you want to add the sig for `fourth` as @ghiculescu suggested before I merge this?

Oops will do now!

PR #279, opened 2020-08-24
Update activerecord.rbi

giovannibonetti:
As [documented in ApiDock](https://apidock.com/rails/v6.0.0/ActiveRecord/ConnectionAdapters/SchemaStatements/index_exists%3F), the method `index_exists?` from `ActiveRecord::ConnectionAdapters::SchemaStatements` can be called with either a single column, or multiple columns in an array:
```ruby
# Check an index exists
index_exists?(:suppliers, :company_id)

# Check an index on multiple columns exists
index_exists?(:suppliers, [:company_id, :company_type])
```


giovannibonetti:
Thanks! 👍

PR #280, opened 2020-08-26
Add `index_with`

ghiculescu:
A new method added in Rails 6 https://blog.bigbinary.com/2019/06/17/rails-6-adds-enumerable-index_with.html


connorshea:
This needs to define the `default` parameter or the RBI will be disabled when you generate RBIs because the parameters will not match between what sorbet autogenerates and what the sorbet-typed definition says it has.

PR #281, opened 2020-09-12
Add a missing param to index_with.

connorshea:
The `index_with` method has a `default` parameter that wasn't added in #280. https://api.rubyonrails.org/classes/Enumerable.html#method-i-index_with

This caused sorbet to mark the file false. As far as I can tell, this should only cause a change in that the file will be marked as strict instead now that it matches the method signature that sorbet finds automatically.

If the user actually passes the `default` parameter to the method, the type returned will be incorrect, but that's not really fixable and I agree with the original logic that it wasn't a likely-enough case that it mattered.

cc: @ghiculescu


ghiculescu:
Thanks, good call.

elliottt (approved):
Thanks!

PR #282, opened 2020-10-10
Type parse-cron

jdkaplan:
I'm starting the types for https://github.com/siebertm/parse-cron. It's not as popular as the other gems in here, but I had fun typing it, and maybe someone else will find it useful.

The library makes some effort to support a few features that make it difficult to type, so I'd especially appreciate feedback on these points:
1. An alternate provider of `Time.now` and `Time.local` (`time_source`). I handled this with generics, but the documentation for those is stil in progress. Are user-facing generics stable enough to be used in this repo?
2. Some methods (`next` and `last`) are overloaded to return different types depending on the value of their arguments. My understanding is that user-facing overloads are not planned. So I chose to type the simpler signature (because that's all I really need) and left the complicated one in as a comment. Would you recommend something different?
3. There are a few tuple types in here. The [tuple docs](https://sorbet.org/docs/tuples) mention that they're experimental. Are they too experimental for shared RBIs like this?

### DIFF
--- a/lib/parse-cron/all/parse-cron.rbi
+++ b/lib/parse-cron/all/parse-cron.rbi
@@ -15,12 +15,12 @@ class CronParser
   def interpret_vixieisms(spec); end
 
   # returns the next occurence after the given date
-  sig { params(now: Time, num: Integer).returns(T.any(Time, T::Array[Time]))  }
-  def next(now = @time_source.now, num = 1); end
+  sig { params(now: Time).returns(Time)  }
+  def next(now = @time_source.now); end
 
   # returns the last occurence before the given date
-  sig { params(now: Time, num: Integer).returns(T.any(Time, T::Array[Time]))  }
-  def last(now = @time_source.now, num=1); end
+  sig { params(now: Time).returns(Time)  }
+  def last(now = @time_source.now); end
 
   SUBELEMENT_REGEX = T.let(%r{}, Regexp)
 
--- a/lib/parse-cron/all/parse-cron.rbi
+++ b/lib/parse-cron/all/parse-cron.rbi
@@ -1,0 +1,134 @@
+# typed: strict
+
+# Parses cron expressions and computes the next occurence of the "job"
+#
+class CronParser
+  extend T::Sig
+  extend T::Generic
+
+  # NOTE: Make these variant if you use the alternate sigs for later methods.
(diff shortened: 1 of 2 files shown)


jdkaplan:
So I don't forget, I'd like this to count as a [Hacktoberfest](https://hacktoberfest.digitalocean.com) contribution. If you merge this, can you please add the `hacktoberfest-accepted` label to this PR?

jez (approved):

lib/parse-cron/all/parse-cron.rbi
```diff
@@ -0,0 +1,134 @@
...
+
+# Parses cron expressions and computes the next occurence of the "job"
+#
+class CronParser
+  extend T::Sig
+  extend T::Generic
+
+  # NOTE: Make these variant if you use the alternate sigs for later methods.
+  SourceClass = type_member(fixed: T.class_of(Time))
+  Source = type_member(fixed: Time)
```
jez, line 11: It looks like you're trying to define a type alias. Why are generics required here?

```suggestion
  SourceClass = T.type_alias {T.class_of(Time)}
  Source = T.type_alias {Time}
```

also it's not clear to me that these type aliases are pulling their weight. Does the parse-cron job actually use Sorbet and define these type_members / type_aliases? If not, by adding them, Sorbet will think that this is ok:

```
puts CronParser::Source
```

when in that's a NameError because that constant doesn't exist at runtime.

`Time` is fewer characters than `Source`. I'd just inline `Time` into all the sigs, and same with `T.class_of(Time)`

jdkaplan: Thanks for reviewing this!

parse-cron doesn't use Sorbet. These generics/aliases are from me trying to fully type this behavior:
```ruby
ExtendedTime = Class.new(Time)
result = CronParser.new("* * * * *", ExtendedTime).next
result.class == ExtendedTime
```
I don't use that at all, though. I was mostly going for completeness (mostly to play with Sorbet).

I didn't realize that Sorbet would think the names given to type members are valid constants. I thought they'd be invisible during type checking outside of that class.

Given all this, I agree that it makes sense to inline them. I'll do that and include a small test file to show that this at least covers the basic cases.

jez (approved):

PR #283, opened 2020-10-21
Fix RBI for ActionDispatch::Routing get + lambda

giovannibonetti:
This is a small fix for ActionDispatch::Routing::Mapper::HttpHelpers#get.

When called with a lambda it requires a parameter, which is `env`.

In my app I have to call it like this:
```ruby
get 'something', to: -> (env) { ... }
```

If no parameter is passed it results in an error:
```ruby
get 'something', to: -> { ... }
```
```
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./config/routes.rb:376:in `block (2 levels) in <main>'
```


elliottt (approved):
Thank you!

PR #291, opened 2020-11-03
Update activerecord.rbi, activemodel.rbi

jeffcarbs:
Update RBI for ActiveRecord and ActiveModel. Also added a few things to the RBI for ActiveSupport that were dependencies of the other two.

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -14,6 +14,48 @@ AssociationCallback = T.type_alias do
   )
 end
 
+class ActiveRecord::ConnectionAdapters::AbstractAdapter; end
+module ActiveRecord::QueryCache::ClassMethods; end
+module ActiveRecord::Querying; end
+module ActiveRecord::Translation; end
+module ActiveRecord::DynamicMatchers; end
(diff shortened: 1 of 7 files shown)


jez (approved):

lib/activerecord/all/activerecord.rbi
```diff
@@ -14,6 +14,48 @@ AssociationCallback = T.type_alias do
...
+module ActiveRecord::Attributes; end
+module ActiveRecord::AttributeDecorators; end
+module ActiveRecord::Locking::Optimistic; end
+module ActiveRecord::Callbacks; end
+module ActiveRecord::Timestamp; end
+module ActiveRecord::Aggregations; end
+module ActiveRecord::NoTouching; end
+module ActiveRecord::Reflection; end
+module ActiveRecord::Serialization; end
+module ActiveRecord::Store; end
```
manhhung741, line 35: @jeffcarbs I feel like we should not add the static module definition and methods without signature to sorbet-typed. Those are usually handled by generated libraries like `rbi gems` or `tapioca`. Adding them to sorbet-typed increases the risk that there are conflicts between different tools.

PR #298, opened 2020-11-12
Fix activerecord sanitize_sql_for_assignment sig

dikond:


elliottt:
I'm not sure why this change is causing the tests to fail, could you take a look?

elliottt:
I think this will pass with a rebase, thanks again for fixing master!

dikond:
@elliottt now CI is 👌

trevor-stripe (approved):
Thanks!

PR #301, opened 2020-11-13
Fix activestorage versions

dikond:
This is to fix CI issues in https://github.com/sorbet/sorbet-typed/pull/299 and https://github.com/sorbet/sorbet-typed/pull/298

You man see that ` lib/activestorage/<=6.1/activestorage.rbi ` violates [version constraints](file:///private/var/folders/b5/l52d7lsn50d4nh32mc7ny7bh0000gn/T/1179.html#version-constraints) and I believe this is the reason was failing.

I also checked when the module in question was added to the activestorage gem. It appeared only in v6.0 and therefore shall be excluded from the `lib/activestorage/all/activestorage.rbi `.


jez (approved):
Thanks!

connorshea:
As I said in the commit message, the reason this was <=6.1 (which I think should be valid?) was because Rails 6.1 adds a new `service` argument to the ActiveStorage methods: https://github.com/sorbet/sorbet-typed/commit/b61b73d1e65546b3331d2fbec23e756f118481aa

Also, this causes problems on Rails 6.1 because it defines `ActiveStorage::Attached::Model::ClassMethods` twice (`>=6.0.0` and `>=6.1` are both pulled in on 6.1). If you want the `>=6.0.0` file to only show up on Rails 6.0.x versions, it should be `~>6.0.0`.

connorshea:
Fixed a few of these issues with #303 :)

dikond:
Good catch with `~>6.0.0`, thanks for fixing it.

PR #303, opened 2020-11-14
ActiveStorage fixes for Rails 6.1

connorshea:
`ActiveStorage::Blob` and `ActiveStorage::Attachment` have a different parent class in Rails 6.1, so that needed to be updated. Also fixed a few issues with the version constraints on a few of these files to make them work correctly with ActiveStorage 6.1.0.rc1.


jez (approved):

connorshea:
Thanks @jez :)

PR #308, opened 2020-11-23
Add RBI for doorkeeper

jeffcarbs:
This was generated via `tapioca generate doorkeeper` at v5.4.0. I set the version to `>=5` since I'm not confident enough to say what came from each version and `v5.0.0` was released over 2 years ago (see: https://rubygems.org/gems/doorkeeper/versions).

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -261,6 +261,9 @@ end
 class ActionController::ActionControllerError < ::StandardError
 end
 
+class ActionController::API < ::ActionController::Metal
+end
+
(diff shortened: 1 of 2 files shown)


lib/actionpack/all/actionpack.rbi
```diff
@@ -261,6 +261,9 @@ end
 class ActionController::ActionControllerError < ::StandardError
 end
 
+class ActionController::API < ::ActionController::Metal
```
jeffcarbs, line 264: The base doorkeeper controller inherits from this.

elliottt (approved):
Thank you!

PR #310, opened 2020-11-28
Adding Gem::Version

beanieboi:
Adding some signatures for Gem::Version


jez (changes requested):
We track types for things in the Ruby standard library inside Sorbet:

https://github.com/sorbet/sorbet/blob/d0119cbdcba3771b493e677f02703481de5f1077/rbi/stdlib/rubygems.rbi#L1519

Can you add the types there?

beanieboi:
that's interesting. This means there are already signatures for Gem::Version. but I get the following error when I enable `strong` typing

```
lib/ffprober/ffmpeg/exec.rb:11: This code is untyped https://srb.help/7018
    11 |      CHAPTER_SUPPORT = T.let(Gem::Version.new('2.0.0'), Gem::Version)
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'll close this PR. Thanks!

PR #311, opened 2020-12-03
Update lib/httparty/all/httparty.rbi

jeffcarbs:
Merge our (Gusto's) hand-maintained httparty.rbi with the upstream.

NOTE: The former RBI was generated via `srb rbi gems` which doesn't automatically figure out things like `mixes_in_class_methods`. The latter RBI was generated via `tapioca generate` which does figure that out. Given that the latter would be more correct I chose to favor that for similar RBI.


elliottt (approved):
Thanks!

PR #312, opened 2020-12-03
Create lib/annotate/all/annotate.rbi

jeffcarbs:
Generated via `tapioca generate annotate` at v3.1.1. I think using `all` here is pretty safe since the main API for this is a rake task and I'd assume most people never interact with the ruby code outside of `Annotate.load_tasks` which had been unchanged since v1.3: https://github.com/ctran/annotate_models/blame/develop/lib/annotate.rb#L74

### DIFF
--- a/lib/annotate/all/annotate.rbi
+++ b/lib/annotate/all/annotate.rbi
@@ -1,0 +1,218 @@
+# typed: strong
+
+module Annotate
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Your argument for `all` makes sense, thanks!

PR #313, opened 2020-12-03
Update lib/paper_trail/all/paper_trail.rbi

jeffcarbs:
Merge our (Gusto's) hand-maintained paper_trail.rbi with the upstream.


elliottt (approved):
Thanks!

PR #314, opened 2020-12-03
Fix sig for rescue_from in active_support.rbi

jeffcarbs:
The block in `rescue_from` has an argument that is the rescued exception: https://api.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html#method-i-rescue_from


elliottt (approved):
Thanks for including the doc link!

PR #319, opened 2020-12-09
Create lib/responders/all/responders.rbi

jeffcarbs:
Generated via `tapioca generate` at v3.0.1. Set the version to `all` here since the main API (`rescue_from` and `rescue_with`) has remained constant for a long while.


elliottt (approved):
Thanks!

PR #321, opened 2020-12-09
Add RBI for AbstractController::Base to actionpack.rbi

jeffcarbs:
The class definition/methods were generated via `tapioca generate`. I ran the generation on all of actionpack but only pulled in this particular class. There are some more diffs b/w this and the full tapioca-generated file but I didn't have time to do the manual merging of all of it.

### DIFF
--- a/lib/actionpack/all/actionpack.rbi
+++ b/lib/actionpack/all/actionpack.rbi
@@ -4,7 +4,45 @@
 module ActionController
 end
 
-class AbstractController::Base < Object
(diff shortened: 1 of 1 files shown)


elliottt:
Was this also generated with tapioca?

jeffcarbs:
@elliottt - This did come from tapioca. I added a PR description explaining the context of the change.

elliottt (approved):
Thanks!

PR #322, opened 2020-12-09
Update lib/railties/all/railties.rbi

jeffcarbs:
Also made some updates to activesupport and activerecord RBI which the updates to railties depended on.

### DIFF
--- a/lib/railties/all/railties.rbi
+++ b/lib/railties/all/railties.rbi
@@ -1,18 +1,1274 @@
 # typed: strong
 
 module Rails
-  sig {returns(Rails::Application)}
-  def self.application; end
-  
-  sig { returns(Pathname) }
-  def self.root; end
+  extend(::ActiveSupport::Autoload)
+
+  class << self
+    def app_class; end
+    def app_class=(_arg0); end
+
+    sig { returns(Rails::Application) }
+    def application; end
+
+    def application=(_arg0); end
+    def autoloaders; end
+
+    sig { returns(ActiveSupport::BacktraceCleaner) }
+    def backtrace_cleaner; end
+
+    sig { returns(ActiveSupport::Cache::Store) }
+    def cache; end
+
+    def cache=(_arg0); end
+    def configuration; end
+
+    sig { returns(ActiveSupport::StringInquirer) }
+    def env; end
+
+    def env=(environment); end
+    def gem_version; end
+    def groups(*groups); end
+    def initialize!(*_arg0, &_arg1); end
+    def initialized?(*_arg0, &_arg1); end
+
+    sig { returns(Logger) }
+    def logger; end
+
+    def logger=(_arg0); end
+    def public_path; end
+
+    sig { returns(Pathname) }
+    def root; end
+
+    sig { returns(String) }
+    def version; end
+  end
 end
 
-class Rails::Application
-  sig {params(blk: T.proc.bind(Rails::Application).void).void}
-  def configure(&blk); end
+class Rails::Application < ::Rails::Engine
+  def initialize(initial_variable_values = T.unsafe(nil), &block); end
 
-  sig {returns(ActionDispatch::Routing::RouteSet)}
+  def asset_precompiled?(logical_path); end
+  def assets; end
+  def assets=(_arg0); end
+  def assets_manifest; end
+  def assets_manifest=(_arg0); end
+  def build_middleware_stack; end
+
+  sig { returns(Rails::Application::Configuration) }
+  def config; end
+
+  def config=(_arg0); end
+  def config_for(name, env: T.unsafe(nil)); end
+  def console(&blk); end
+  def credentials; end
+  def default_url_options(*_arg0, &_arg1); end
+  def default_url_options=(arg); end
+  def eager_load!; end
+  def encrypted(path, key_path: T.unsafe(nil), env_key: T.unsafe(nil)); end
+  def env_config; end
+  def executor; end
+  def generators(&blk); end
+  def helpers_paths; end
+  def initialize!(group = T.unsafe(nil)); end
+  def initialized?; end
+  def initializer(name, opts = T.unsafe(nil), &block); end
+  def initializers; end
+  def isolate_namespace(mod); end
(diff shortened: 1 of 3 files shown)


jez (approved):

lib/railties/all/railties.rbi
```diff
@@ -1,18 +1,1274 @@
...
+    sig { returns(ActiveSupport::BacktraceCleaner) }
+    def backtrace_cleaner; end
+
+    sig { returns(ActiveSupport::Cache::Store) }
+    def cache; end
+
+    def cache=(_arg0); end
+    def configuration; end
+
+    sig { returns(ActiveSupport::StringInquirer) }
```
connorshea, line 25: While technically more accurate, I'd argue this is a bad change because now `Rails.env.development?` and similar methods do not pass typechecking even though they should.

@jeffcarbs do you have a fix for this that you use, or are any files that use `Rails.env.development?` just marked false?

lib/railties/all/railties.rbi
```diff
@@ -1,18 +1,1274 @@
...
+  def initialize(initial_variable_values = T.unsafe(nil), &block); end
+
+  def asset_precompiled?(logical_path); end
+  def assets; end
+  def assets=(_arg0); end
+  def assets_manifest; end
+  def assets_manifest=(_arg0); end
+  def build_middleware_stack; end
+
+  sig { returns(Rails::Application::Configuration) }
```
connorshea, line 58: I'm also opposed to this change for similar reasons as the others. The problem is that there are a _ton_ of configuration options in Rails, so while this is technically accurate, it means that any files that use `Rails.application.config.whatever` are going to now be marked `false` and fail to typecheck.

Unless we mapped every config method in rails (I wouldn't recommend it), I think this should remain untyped.

lib/railties/all/railties.rbi
jeffcarbs: @connorshea - we have the following RBI in a “manual” RBI file:
```
class ActiveSupport::StringInquirer
  sig { returns(T::Boolean) }
  def development?; end

  sig { returns(T::Boolean) }
  def production?; end

  sig { returns(T::Boolean) }
  def staging?; end

  sig { returns(T::Boolean) }
  def test?; end
end
```

I do see how this could be annoying though. If you want to remove this sig and maybe leave a comment as to why it’s omitted I won’t be offended 🙂 I can just add it as RBI in my repo.

lib/railties/all/railties.rbi
jeffcarbs: We have a similar manual RBI in our repo where we type the configuration files with the custom methods we define.

Similar thing though, I’d be cool if you want update this to be unsigged with a comment explaining why.

lib/railties/all/railties.rbi
connorshea: hmm, maybe add those signatures as a comment in the railties.rbi above the current `env` signature? That way we can lead people toward what they _should_ be doing? I definitely like having the methods typecheck correctly and your solution seems pretty good.

jeffcarbs: I’ve done that in other RBI where you can’t really make assumptions about other people’s setup. For instance, sidekiq defines rails extensions but they’re only relevant if you’re using rails. If you typed them in the sidekiq.rbi it would incorrectly tell sorbet ActiveRecord was there. For example: https://github.com/sorbet/sorbet-typed/blob/master/lib/sidekiq/all/sidekiq.rbi#L102-L105

lib/railties/all/railties.rbi
jeffcarbs: We have a custom config attribute `Rails.application.config.redis_manager` so we have the following RBI:
```
class Rails::Application::Configuration
  def redis_manager; end
  def redis_manager=(value); end
end
```

connorshea: I'd rather push it to the consumer, just because otherwise it causes type errors from Sorbet not understanding what `Rails.application.config` can be chained with

lib/railties/all/railties.rbi
ghiculescu: This returns an `ActiveSupport::EnvironmentInquirer` from Rails 6.1. See https://github.com/rails/rails/pull/38657

We should return the correct type in those cases, but we should also do this:

```ruby
class ActiveSupport::EnvironmentInquirer
  sig { returns(T::Boolean) }
  def development?; end

  sig { returns(T::Boolean) }
  def production?; end

  sig { returns(T::Boolean) }
  def test?; end
end
```

ghiculescu: in my opinion it's also okay to ship this to everyone:

```ruby
class ActiveSupport::StringInquirer
  sig { returns(T::Boolean) }
  def development?; end

  sig { returns(T::Boolean) }
  def production?; end

  sig { returns(T::Boolean) }
  def staging?; end

  sig { returns(T::Boolean) }
  def test?; end
end
```

technically, *any* method that ends with `?` when called on an `ActiveSupport::StringInquirer` will return a `T::Boolean`. unlike the sidekiq example this doesn't introduce new incorrect dependencies.

jeffcarbs: Ah you're right, it will technically never raise
```ruby
> Rails.env.womp?
=> false
```

I think this is one of those places where the RBI can enforce things that aren't enforced at runtime. For example, while the string/environment inquirer technically won't raise if you ask about a bad env, it would be nice for sorbet to catch the typo. Otherwise you might think you're checking for a particular environment when you actually aren't 😬 

How about this: since `production` `test` and `development` are default environments (i.e. created when you `rails new`), we can ship RBI for them. So:

```
class ActiveSupport::EnvironmentInquirer # And StringInquirer depending on Rails version
  sig { returns(T::Boolean) }
  def development?; end

  sig { returns(T::Boolean) }
  def production?; end

  sig { returns(T::Boolean) }
  def test?; end
end
```

I think we can leave a comment for people to add other environments (e.g. `staging`) if they need them. We actually have a few other environments so would need to do that anyway.

PR #323, opened 2020-12-09
Create lib/net-sftp/all/net-sftp.rbi

jeffcarbs:
Generated via `tapioca generate` at v3.0.0.

The RBI was removed from sorbet proper in https://github.com/sorbet/sorbet/pull/3516 so adding the RBI to sorbet-typed.


elliottt (approved):
Thank you for closing the loop on this one! :)

PR #330, opened 2021-01-05
Add YARD

dduugg:
The [autogenerated](https://github.com/dduugg/yard-sorbet/blob/3f0506a/sorbet/rbi/gems/yard.rbi) `YARD` rbi for a gem I maintain wasn't detecting `YARD::CodeObjects::ClassObject`. Similarly, the [autogenerated](https://github.com/dduugg/yard-sorbet/blob/3f0506a/sorbet/rbi/hidden-definitions/hidden.rbi#L11100-L11112) hidden-definitions file wasn't detecting the inheritance or `initilialize` method of the class.

This is a blocker to typing [one](https://github.com/dduugg/yard-sorbet/blob/8ecb9f0/lib/yard-sorbet/struct_handler.rb#L55) of the files in the gem.

Thus, I've started with the autogenerated YARD stub (first commit), and added code to correctly type the `ClassObject` initializer (second commit). There's a small test as well (third commit).

The class in YARD source code is [here](https://github.com/lsegal/yard/blob/main/lib/yard/code_objects/class_object.rb#L7).


elliottt (approved):
Thanks @dduugg!

PR #332, opened 2021-01-13
Add rspec-core

dduugg:
Adds `rspec-core`, in order to pick up `RSpec::Core::RakeTask`, which isn't inferred by `srb`, but is a [common](https://github.com/search?q=RSpec%3A%3ACore%3A%3ARakeTask&type=code) incantation.

1st commit: adds a test
2nd commit: adds the autogenerated `rspec-core`
3rd commit: adds `RSpec::Core::RakeTask`

One potential hiccup is that the class inherits `::Rake::TaskLib`, which isn't found in this repo, or in the autogenerated `Rake` code, but doesn't seem to present issues (tests pass locally as well as in a project that uses `RSpec::Core::RakeTask`).

PR #333, opened 2021-01-13
Add Rake

dduugg:
Adds Rake, specifically to pick up `Rake::TaskLib` and un-block https://github.com/sorbet/sorbet-typed/pull/332


jez (approved):
ty

PR #344, opened 2021-03-01
has_one allows touch as a parameter

dgodd:


lib/activerecord/all/activerecord.rbi
```diff
@@ -151,6 +152,7 @@ module ActiveRecord::Associations::ClassMethods
     source: nil,
     source_type: nil,
     through: nil,
+    touch: nil,
     validate: nil,
```
jez, line 156:
```suggestion
    validate: nil,
    touch: nil,
```

dgodd:
Thanks for the suggestion @jez

PR #347, opened 2021-03-20
Remove untyped methods from activerecord.rbi

connorshea:
No reason to have all these methods, they only make maintaining forward/backward compatibility more difficult for no benefit.

If you want to know all the methods that exist in ActiveRecord, you should be generating gem RBIs with Sorbet or Tapioca.

I've left in mixes_in_class_methods usages, to make sure that this commit doesn't remove anything that `srb rbi gems` wouldn't know about itself.

This resolves a lot of issues with the RBI (e.g. different arguments between these sigs and the actual methods in rails).

### DIFF
--- a/lib/activerecord/all/activerecord.rbi
+++ b/lib/activerecord/all/activerecord.rbi
@@ -35,25 +35,7 @@ module ActiveRecord::Serialization; end
 module ActiveRecord::Store; end
 
 module ActiveRecord::Associations
-  extend(::ActiveSupport::Concern)
-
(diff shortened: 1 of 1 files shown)


jez:
I'm fine to land this but can you ask in the #rails channel on Slack and confirm whether this is something most people want?

manhhung741 (approved):
I like this because there are tools for generating method definition without signatures like tapioca and srb rbi gems

jez (approved):

PR #349, opened 2021-04-07
Update ActiveRecord's accepts_nested_attributes_for signature

robotdana:
- add the attributes param to the reject_if proc
- allow the proc arguments to be Proc.new. 
- allow multiple associations to be specified
- add accepts_nested_attirbutes_for to the test file


froydnj (approved):

froydnj:
Thanks!

PR #354, opened 2021-05-28
rm autogenerated rubocop components

dduugg:
Reduces https://github.com/sorbet/sorbet-typed/pull/334 to just the part that sorbet doesn't autogenerate.

This otherwise creates conflicts with the contents of `sorbet/rbi/gems/rubocop.rbi` in subsequent `rubocop` releases. (I assumed this `rbi` would replace, not co-exist with, the autogenerated rbi. It might be worth clarifying this in the docs.)


elliottt (approved):
Thanks dug!

PR #357, opened 2021-06-19
Fix type signatures for after_*_commit in ActiveRecord.

connorshea:
Ran into this when trying to add Sorbet to a larger Rails app.

These methods takes a block and the first argument actually accepts multiple values. Updated the tests accordingly.

See the docs: https://api.rubyonrails.org/v5.2.0/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_create_commit


jez (approved):

connorshea:
Thanks @jez!

PR #358, opened 2021-06-19
Add more methods to the aasm gem and improve the existing sigs.

connorshea:
And add tests for the aasm gem.

This comes from messing with adding Sorbet to a large Rails app that uses aasm. There were a number of problematic/incorrect type signatures in the RBI file that existed before.


elliottt (approved):
Thank you!

PR #359, opened 2021-06-22
Misc improvements

connorshea:
Fixes for a few different issues I've discovered trying to add Sorbet to a large Rails app.

- `after_*`, `before_*`, etc. callbacks accept more than one argument, which the types weren't handling correctly before.
- `T.proc.void` binds to the parent class rather than an instance of the class, and I couldn't figure out a way to implement this in an accurate way, so I've just made it bind to untyped instead.
- `saved_change_to_attribute?` accepts a string as well as a symbol.

### DIFF
--- a/lib/activerecord/>=5/activerecord.rbi
+++ b/lib/activerecord/>=5/activerecord.rbi
@@ -7,7 +7,7 @@ class ActiveRecord::Base
       args: T.any(Symbol, T.proc.returns(T.untyped)),
       if: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean)))),
       unless: T.nilable(T.any(Symbol, Proc, T.proc.params(arg0: T.untyped).returns(T.nilable(T::Boolean)))),
-      block: T.nilable(T.proc.void)
+      block: T.nilable(T.proc.bind(T.untyped).void)
     ).void
   end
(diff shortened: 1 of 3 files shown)


elliottt (changes requested):
Shopify added `T.bind` recently, which allows you to specify what the type of `self` will be inside of a block passed as a normal argument. Maybe that would help in avoiding the use of `T.proc.bind(T.untyped)`?

https://sorbet.org/docs/type-assertions#tbind

connorshea:
@elliottt that can be used to resolve the problem on a case-by-case basis, but I'd like it to not error by default in this case since it makes adoption of Sorbet in files using these methods more difficult. As far as I know, this change isn't incompatible with using T.bind in the individual blocks to improve the type accuracy of the given block, so I don't think this PR would be a problem?

elliottt:
We had been discussing not allowing `.bind` to occur on proc-typed parameters other than those used as the block parameter of a method, so it's possible that this could need to be reverted in the future.

elliottt:
> We had been discussing not allowing `.bind` to occur on proc-typed parameters other than those used as the block parameter of a method, so it's possible that this could need to be reverted in the future.

I'm not sure how I missed that these were to the block parameter of the methods -- I had read them as being additional parameters of proc-typed arguments. Looks good!

elliottt (approved):

PR #361, opened 2021-07-05
Add Parlour 6.0.1 RBI

AaronC81:
Parlour has just been updated to 6.0.1, so this PR adds the new RBI for this version.

(Nothing has changed since the previous version, besides the version number constant!)


froydnj (approved):
Thanks!

PR #365, opened 2021-07-22
More stripe types

iloveitaly:
I'm not sure how to represent `period` correctly.

Returning StripeObject won't work since the returning object contains start/end, which StripeObject does not.

```
so = Stripe::StripeObject.construct_from({"start": 1626978473, "end": 1626978473})
so.start
so.end
```


froydnj (approved):
Thanks!

PR #372, opened 2021-09-13
Fix invalid initialize sigs

dduugg:
Just a pair of sketchy `perl` expressions that attempt to fix the sigs that are invalid as of https://github.com/sorbet/sorbet/pull/4533

```sh
perl -p -i -e 's/returns\S+(\s?\}\n\s*def initialize)/void$1/g' -0777 **/*.rbi
perl -p -i -e 's/returns\S+(\n\s*end\n\s*def initialize)/void/g' -0777 **/*.rbi
```

I assume y'all came up with something better internally, feel free to discard this in favor of it.

(Gentle plea to please consider adding this repo to the `sorbet` test suite)


elliottt (approved):
Thanks so much for fixing these @dduugg!

PR #374, opened 2021-10-05
Add YARD::CodeObjects::Proxy

dduugg:
see https://github.com/lsegal/yard/blob/main/lib/yard/code_objects/proxy.rb#L24


jez (approved):

jez (changes requested):
looks like the test you added doesn't type check

dduugg:
How embarrassing 🙈 , tests fixed.

I also added another commit to remove all the unnecessary autogenerated components, see https://github.com/sorbet/sorbet-typed/pull/354 for rationale. Apologies for the scope creep, lmk if you would prefer a separate PR.

PR #380, opened 2021-10-20
Fixing sidekiq return type

iloveitaly:


iloveitaly:
@jez fixed. I need to debug why I can't run these tests locally and get descriptions of what is wrong.

PR #383, opened 2021-11-12
Add more options to aasm event

tanzoniteblack:
See https://github.com/aasm/aasm/blob/master/lib/aasm/core/event.rb#L20 for the options I'm adding here.


tanzoniteblack:
Closing this for the moment because these should actually be lists of symbols

PR #384, opened 2021-11-16
Add additional parameters to aasm event

tanzoniteblack:
The list of the options I'm adding can be found [here](https://github.com/aasm/aasm/blob/master/lib/aasm/core/event.rb#L20).

Also updates `after_commit` and `guard` to be nillable arrays of Symbols instead of just nilable symbols.


aprocter-stripe (approved):
Thanks!

PR #385, opened 2021-11-30
rm autogenerated rspec-core components

dduugg:
Analogous to https://github.com/sorbet/sorbet-typed/pull/354 but for `rspec-core`

Should fix https://stackoverflow.com/questions/70172188/sorbet-setup-results-in-tc-complaining-about-rbi-file


jez (approved):

BWStearns:
@dduugg thanks a ton for fixing this so fast!

PR #388, opened 2022-01-11
Re-add definitions for AR::Migration[5.0]

mikker:
Somehow these got removed and I was getting errors for a missing `add_column` definition. Adding this line fixed the problem.


aprocter-stripe (approved):
Thanks!

PR #406, opened 2024-02-29
Update Payment Intent typings for Stripe

munaz2-stripe:


jez (approved):
Note that sorbet-typed is deprecated in favor of rbi-central, because tapioca is the recommended way to manage Sorbet projects and RBIs:

https://github.com/sorbet/sorbet-typed/blob/master/README.md

